PR #288, opened 2020-07-14
Stream: bump amazon-kinesis-client to 2.2.12

chuwy:


config/config.hocon.sample
```diff
@@ -96,7 +100,9 @@ enrich {
       #   creds = ${?GOOGLE_APPLICATION_CREDENTIALS}
       # }
 
-      # Maximum number of records to get from Kinesis per call to GetRecords
+      # Enhanced fan out (dedicated read throughput) vs. polling mode (shared read throughput):
+      # This is the Maximum number of records to get from Kinesis per call to GetRecords when in polling mode.
+      # The default behavior is using enhanced fan-out and it doesn't use this configuration.
```
benjben, line 105:
```suggestion
      # If specified, KCL will use polling mode (shared read throughput) and will get max maxRecords from Kinesis per call to GetRecords.
      # If not specified, KCL will use enhanced fan out (dedicated read throughput).
```

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
```diff
@@ -147,92 +169,149 @@ class KinesisSource private (
...
+      .map(InitialPositionInStreamExtended.newInitialPositionAtTimestamp(_))
+      .getOrElse(InitialPositionInStreamExtended.newInitialPosition(positionValue))
+
+    val metricFactory = kinesisConfig.disableCloudWatch match {
+      case Some(true) => new NullMetricsFactory()
+      case _ => null // KCL internally creates it.
     }
 
-    worker.run()
+    //Enhanced fan-out is the default retrieval behavior for KCL 2.x. So we have to override it.
```
benjben, line 217:
```suggestion
    //Enhanced fan-out is the default retrieval behavior for KCL 2.x. We need to override it if we want to keep using polling mode.
```

oguzhanunlu:
there are unformatted files per scalafmt, incl. `KinesisSource.scala`, `utils.scala`, `BuildSettings.scala`

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
```diff
@@ -91,17 +110,20 @@ class KinesisSource private (
   config: StreamsConfig,
   kinesisConfig: Kinesis,
   sentryConfig: Option[SentryConfig],
-  provider: AWSCredentialsProvider
+  sourceProvider: AwsCredentialsProvider,
+  sinkProvider: AWSCredentialsProvider
```
oguzhanunlu, line 114: Is there a specific reason not to migrate both consumer and producer?

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
```diff
@@ -91,17 +110,20 @@ class KinesisSource private (
...
 ) extends Source(client, adapterRegistry, enrichmentRegistry, processor, config.out.partitionKey, sentryConfig) {
 
   override val MaxRecordSize = Some(1000000)
+  private val DYANMODB_DEFAULT_INITIAL_RCU = 10
+  private val DYANMODB_DEFAULT_INITIAL_WCU = 10
```
oguzhanunlu, line 119:
```suggestion
  private val DYNAMODB_DEFAULT_INITIAL_RCU = 10
  private val DYNAMODB_DEFAULT_INITIAL_WCU = 10
```

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
chuwy: Good point. I think even if there's a reason - I'd like to have it documented somewhere because it feels a bit strange.

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
chuwy: Nice catch

modules/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sources/KinesisSource.scala
```diff
@@ -29,13 +29,30 @@ import scala.util.control.NonFatal
...
+import software.amazon.awssdk.services.dynamodb.DynamoDbAsyncClient
+import software.amazon.awssdk.services.cloudwatch.CloudWatchAsyncClient
+import software.amazon.awssdk.services.kinesis.KinesisAsyncClient
+import software.amazon.awssdk.regions.Region
+import software.amazon.kinesis.common.ConfigsBuilder
+import software.amazon.kinesis.common.KinesisClientUtil
+import software.amazon.kinesis.coordinator.Scheduler
+import software.amazon.kinesis.exceptions.InvalidStateException
+import software.amazon.kinesis.exceptions.ShutdownException
+import software.amazon.kinesis.lifecycle.events.InitializationInput
```
chuwy, line 50: Would like to group some of these imports. Feels unnecessary verbose.

modules/stream/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/utils.scala
```diff
@@ -78,23 +86,54 @@ object utils {
...
+        case AWSCredentials(a, s) if isDefault(a) || isDefault(s) =>
+          "accessKey and secretKey must both be set to 'default' or neither".asLeft
+        case AWSCredentials(a, s) if isIam(a) && isIam(s) =>
+          InstanceProfileCredentialsProvider.create().asRight
+        case AWSCredentials(a, s) if isIam(a) && isIam(s) =>
+          "accessKey and secretKey must both be set to 'iam' or neither".asLeft
+        case AWSCredentials(a, s) if isEnv(a) && isEnv(s) =>
+          EnvironmentVariableCredentialsProvider.create().asRight
+        case AWSCredentials(a, s) if isEnv(a) || isEnv(s) =>
+          "accessKey and secretKey must both be set to 'env' or neither".asLeft
```
oguzhanunlu, line 129: Duplicate code exists in `getAWSCredentialsProvider` and this method.

If we are going to migrate both consumer and producer, this comment isn't valid. Otherwise, we need to deduplicate this snippet.

chuwy:
Pushing back to 1.5.0, as we're short on time on 1.4.0 and KCL 2 will require extensive testing.

chuwy:
Rebased it atop of `develop`. Also while we were waiting KCL 2.3.1 has been released with lots of important improvements - we should use it instead of 2.2.12.

AlJohri:
hey @chuwy, we're running into a few strange issues around the KCL with our enrich deployment and wonder if upgrading to KCL 2.x would help. any updates around getting this PR merged in? thank you!

AlJohri:
I see there is also a version 1.14.1 now released which may be easier to upgrade to for the timebeing: https://github.com/awslabs/amazon-kinesis-client/blob/v1.x/CHANGELOG.md#latest-release-1141---january-27-2021

PR #292, opened 2020-07-16
Common: Skip sql enrichment on null values as placeholders (close #291)

### LINKED ISSUE #291: Common: skip SQL enrichment on null values as placeholders

lukeindykiewicz:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/DbExecutor.scala
```diff
@@ -363,7 +363,7 @@ object DbExecutor {
       case Right(placeholderCount) =>
         placeholderMap match {
           case Some(intMap) if intMap.keys.size == placeholderCount => ().asRight
-          case _ => ().asRight
```
chuwy, line 366: 🤦

Issue #301, opened 2020-07-29
Common: decode schema key before processing in toRawEvents function in Iglu Adapter


chuwy:
We're getting bad rows, produced by Iglu Adapter with following payload:

```
failure={timestamp=2020-07-08T02:54:17.062Z, vendor=com.snowplowanalytics.iglu, version=v1, messages=[{field=schema, value=iglu%3Acom.acme%2Fevent%2Fjsonschema%2F1-0-1, expectation=INVALID_IGLUURI, error=null, json=null, actual=null, expectedmapping=null}]}
```

Value is `iglu%3Acom.acme%2Fevent%2Fjsonschema%2F1-0-1`, while should be `iglu:com.acme/event/jsonschema/1-0-1`. Expectation is `INVALID_IGLUURI`, while it should be in `error`.

chuwy:
False alarm. The incoming data was URL-encoded twice.

PR #305, opened 2020-07-30
Add test to demonstrate bug

dilyand:
A URL-encoded schema URI, as the one we have in `GET` requests via the Iglu Webhook Adapter, results in a bad row instead of being successfully processed.

This test must fail when https://github.com/snowplow/enrich/issues/301 is addressed.


chuwy:
Closing. There was no bug (details in #301)

PR #312, opened 2020-08-11
Skip the SQL enrichment on null placeholders

lukeindykiewicz:
This actually make the enrichment to be skipped, the previous one just fixed the one function.

### DIFF
--- a/modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/sqlquery/SqlQueryEnrichmentIntegrationTest.scala
+++ b/modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/sqlquery/SqlQueryEnrichmentIntegrationTest.scala
@@ -22,8 +22,6 @@ import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer, SelfDescribingData
 import org.specs2.Specification
 
 import outputs.EnrichedEvent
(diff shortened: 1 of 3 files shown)


modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/sqlquery/SqlQueryEnrichmentIntegrationTest.scala
```diff
@@ -407,12 +405,7 @@ class SqlQueryEnrichmentIntegrationTest extends Specification {
     val config = SqlQueryEnrichment.parse(configuration, SCHEMA_KEY).map(_.enrichment)
     val context = config.toEither.flatMap(_.lookup(event, Nil, Nil, None).toEither)
 
-    context must beLeft.like {
-      case nel =>
-        nel.head.asInstanceOf[EnrichmentFailure].message must beEqualTo(
-          FailureDetails.EnrichmentFailureMessage.Simple("The placeholder map error. The map: Some(IntMap()), where count is: 1")
-        )
-    }
+    context ==== Right(Nil)
```
chuwy, line 408: What's the difference with `beRight(Nil)`?

lukeindykiewicz: The `====` is safe in term of comparing different types - it doesn't compile. The `must beRight` isn't.

PR #326, opened 2020-08-27
Cats 2.x

chuwy:
- [x] scala-weather
- [x] scala-forex
- [x] scala-referer-parser


chuwy:
Hey @benjben! Can you please review this one? It still technically not done since we're still missing forex, weather and referer-parser, but those will be one line changes, while YAUAA update is relatively big here.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala
```diff
@@ -63,14 +84,19 @@ object YauaaEnrichment extends ParseableEnrichment {
...
 }
 
 /**
  * Class for YAUAA enrichment, which tries to parse and analyze the user agent string
  * and extract as many relevant attributes as possible, like for example the device class.
+ * @param model version of yauaa_context produced, `1` and `2` are supported currently
  * @param cacheSize Amount of user agents already parsed that stay in cache for faster parsing.
  */
-final case class YauaaEnrichment(cacheSize: Option[Int]) extends Enrichment {
+final case class YauaaEnrichment(model: Int, cacheSize: Option[Int]) extends Enrichment {
```
benjben, line 99:
```suggestion
final case class YauaaEnrichment(cacheSize: Option[Int], model: Int = YauaaEnrichment.DefaultContextModel) extends Enrichment {
```

benjben (approved):
This looks great @chuwy :+1:

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala
chuwy: I prefer to avoid defalut parameters as much as possible. In libs they're 100% a binary-compatibility breakage in apps often cause problems with type class derivation.

benjben: I see, good to know!

chuwy: On the other hand FYI, in this case we'd already broke the binary compatibility by adding a parameter to case class: https://github.com/jatcwang/binary-compatibility-guide#avoid-using-case-classes

chuwy: ...although case class parameter is also source incompatibility, so at least its very explicit.

benjben: Thanks for the link, very interesting!

oguzhanunlu (approved):

chuwy:
Force-pushed too fix YAUAA test. Merging.

Issue #56, opened 2020-06-19
Common: Loader.toCollectorPayload should emit only one bad row and not a list


chuwy:
```scala
Loader[T] {
  def toCollectorPayload(
    line: T,
    processor: Processor
  ): ValidatedNel[BadRow.CPFormatViolation, Option[CollectorPayload]]
```

The function signature indicates that several `CPFormatViolation` bad rows can be emitted for one `CollectorPayload`, whereas it should be maximum one.

chuwy:
Migrated from https://github.com/snowplow/snowplow/issues/4315 (comments are auto-generated)

chuwy:
Following this conversation: https://github.com/snowplow/enrich/pull/328#discussion_r489459966, @benjben. I checked this function and in fact it's not that easy to get rid of `NonEmptyList` right now - we're blocked by https://github.com/snowplow-incubator/snowplow-badrows/issues/51

PR #328, opened 2020-09-01
Stream Enrich: New Hope

chuwy:
- [x] Make `AssetsRefresh` respect non-2xx responses (retry or give up)
- [x] Logging
- [x] `blockOn` in `enrichWith`
- [x] More tests
- [x] ack
- [x] Sentry
- [x] File sink
- [x] CI/CD
- [x] Merge #326 into `develop` and rebase on top of it
- [x] Metrics
- [x] Extract non-NH commits and move to #364 
- [x] Make sure `0` in assets/metrics periods counts as "disabled"


modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Main.scala
```diff
@@ -0,0 +1,41 @@
...
+import cats.syntax.flatMap._
+
+import cats.effect.{Blocker, Resource, ExitCode, IOApp, IO}
+
+object Main extends IOApp {
+  def run(args: List[String]): IO[ExitCode] =
+    EnrichJobConfig.command.parse(args) match {
+      case Right(cfg) =>
+        val environment = for {
+          blocker <- Blocker.apply[IO]
```
benjben, line 24: Why are we initializing `blocker` here rather than [here](https://github.com/snowplow/enrich/pull/328/files#diff-b020b3f46b164eb407a4c7553f253befR57-R68) ?

chuwy: No reason I think - you're probably right, it look better in `Environment`.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Main.scala
```diff
@@ -0,0 +1,36 @@
...
+  def run(args: List[String]): IO[ExitCode] =
+    EnrichJobConfig.command.parse(args) match {
+      case Right(cfg) =>
+        for {
+          environment <- Environment.init[IO](cfg).value
+          exit <- environment match {
+            case Right(e) => e.use { env =>
+              val enrich = Enrich.run[IO](env)
+              val updates = AssetsRefresh.run[IO](env)
+              enrich.merge(updates).compile.drain.as(ExitCode.Success)
```
benjben, line 27: Will this `.as(ExitCode.Success)` ever be used?

chuwy: Only if our `Stream` ends... which will never happen, in other words - no.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
```diff
@@ -0,0 +1,77 @@
...
+ */
+case class Payload[F[_], A](data: A, ack: F[Unit]) {
+
+  /**
+   * Flatten all payloads from a list and replace an `ack` action to no-op everywhere
+   * except last message, so that original collector payload (with multiple events)
+   * will be ack'ed only when last event has sunk into good or bad sink
+   */
+  def decompose[L, R](implicit ev: A <:< List[Validated[L, R]], F: Applicative[F]): Stream[F, Parsed[F, L, R]] = {
+    val _ = ev
```
benjben, line 41: What does this line do?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -0,0 +1,132 @@
...
+    val result = Logger[F].debug(payloadToString(payload)) *>
+      Clock[F]
+        .realTime(TimeUnit.MILLISECONDS)
+        .map(millis => new DateTime(millis))
+        .flatMap { now =>
+          EtlPipeline.processEvents[F](adapterRegistry, enrichRegistry, igluClient, processor, now, payload)
+        }
+        .map(enriched => Payload(enriched, row.ack))
+
+    result.handleErrorWith(sendToSentry[F](row, sentry))
```
benjben, line 89: Shouldn't we send the `payload` rather than the `row` ? So that to troubleshoot we don't need to Thrift deserialize it

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
chuwy: Just "uses" `ev`. Otherwise scalac compiles that there's an argument never used.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
chuwy: I think `sendToSentry` is probably slightly misleading. It does three things:

1. Sends an exception to Sentry (we cannot send anything from an event because it can contain PII data
2. Creates a `generic_error` bad row - that's why we need a `row`
3. Logs an error

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
benjben: If it's not used apart from this line, why does the implicit arg need to be there?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
benjben: My point is, why not use `payload` instead of `row` in the generic bad row created? An array of Thrift bytes is not very useful to troubleshoot, compared with a `BadRow` (CPFormatViolation) or a `CollectorPayload`

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
chuwy: It's an compile-time evidence that our `A` is `List[Validated[L, R]]`, so we can use `mapWithLast` on it, which works only with `List`. And our `use` works only with `Validated[L, R]`.

In other words, if `A` is something else, e.g. `Array[Byte]` you won't be able to call `decompose` - it won't compile.

benjben: I see, thanks for the explanation! So in `Payload[A]` `A` can be anything but if we want to use `decompose` A needs to be a `List[Validated[L, R]`, is that right?

Weird that Scala forces us to write `val _ =`

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
chuwy: Ah, ok makes sense now!

The reason was that our payload is something like `ValidatedNel[CPFormatViolation, Option[CollectorPayload]]`, so we don't really have a parsed payload yet. We technicall can pattern-match on it, like:

```scala
payload match {
  case Validated.Invalid(errors) => // what to do here? why we were trying to process error in a first place
    errors // thrift bytes anyway
  case Validated.Valid(payload) =>
    turnIntoAdapterFailure(payload) // but what if it's not an adapter failure? enrichment failure? then we need a raw event to construct bad row
```

And also feels weird to produce different kinds of bad row from the same place, so I decided to stick with the most generic one.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
chuwy:
```
case class Wapper[A](a: A) {
  /** Exists only if its an Int */
  def square(implicit ev: A <:< Int): Int =
    a * a

  def capitalize(implicit ev: A <:< String): String =
    a.capitalize
}
```

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
benjben: I was thinking of putting something like `show"$payload"` in the `generic_error`, whatever it contains.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
chuwy: >  So in Payload[A] A can be anything but if we want to use decompose A needs to be a List[Validated[L, R], is that right?

That's right, yeah.

> Weird that Scala forces us to write val _ =

It's just a value in the end. Moreover we still can use it - it has `apply[A](a: A): List[Validated[L, R]]` method. But I noticed that people prefer `asInstanceOf` in this case.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
benjben: It's actually a [bug](https://github.com/snowplow/enrich/issues/56) that `ThriftLoader.toCollectorPayload`'s signature says that it can return multiple bad rows, it can return only one.

So in the `generic_error` bad row we could put either the single `CPFormatViolation` (recoverable) or the raw extracted `CollectorPayload` (recoverable).

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
```diff
@@ -0,0 +1,289 @@
...
+        val init = for {
+          curDir <- getCurDir
+          _ <- Logger[F].info(s"Initializing assets refresh stream in $curDir")
+          assets <- env.enrichments.get.map(_.configs.flatMap(_.filesToCache))
+          _ <- download(env.blocker, env.assetsState, curDir, assets)
+          _ <- env.pauseEnrich.set(false)
+        } yield updateStream[F](env.blocker, env.assetsState, env.enrichments, curDir, duration, assets)
+        Stream.eval(init).flatten
+      case None =>
+        Stream.empty.covary[F]
```
benjben, line 150: Don't we want to download the assets once for good even if update period is not specified ?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  /** Pair of a tracked `URI` and destination path on local FS (`java.nio.file.Path` is not serializable) */
+  type Asset = (URI, String)
+
+  /** Initialise the [[updateStream]] with all necessary resources if refresh period is specified */
+  def run[F[_]: ConcurrentEffect: ContextShift: Timer: Parallel](env: Environment[F]): Stream[F, Unit] =
+    env.assetsUpdatePeriod match {
+      case Some(duration) =>
+        val init = for {
```
benjben, line 141: Don't we want to pause enrich here?

benjben (commented):
Looks great @chuwy !

I didn't see `env.pauseEnrich.set(false)`, when does this happen?

build.sbt
```diff
@@ -183,3 +183,52 @@ lazy val beam =
...
+  .dependsOn(common)
+  .settings(BuildSettings.basicSettings)
+  .settings(BuildSettings.formatting)
+  .settings(BuildSettings.scoverageSettings)
+  .settings(BuildSettings.addExampleConfToTestCp)
+  .settings(BuildSettings.sbtAssemblySettings)
+  .settings(
+    name := "fs2-enrich",
+    description := "High-performance streaming Snowplow Enrich job built on top of functional streams",
+    buildInfoKeys := Seq[BuildInfoKey](organization, name, version, "sceVersion" -> version.value, description),
```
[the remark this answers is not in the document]
chuwy: It's also odd to see `stream-enrich-1.3.1-common-1.3.1` in `v_etl`, but I thought somebody outside Snowplow can implement their own enrich asset, but now thinking that chances are super low and only one version should be enough. Will delete this one as well.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Main.scala
```diff
@@ -0,0 +1,68 @@
...
+                    case Left(error) =>
+                      logger.error(s"Cannot initialise enrichment resources\n$error").as(ExitCode.Error)
+                  }
+        } yield exit
+      case Left(error) =>
+        IO(System.err.println(error)).as(ExitCode.Error)
+    }
+
+  /** Last attempt to notify about an exception (possibly just interruption) */
+  private def unsafeSendSentry(error: Throwable, sentry: Option[SentryClient]): Unit = {
```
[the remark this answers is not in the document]
chuwy: I made it impure because there's always a super tiny chance of something wrong going with cats-effect infrastructure and I wanted to get rid of everything on the way and maximize the chance of delivering the exception to sentry.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Payload.scala
```diff
@@ -0,0 +1,77 @@
...
+   * Despite of the result (`A` or `B`) the original one still has to be acknowledged
+   *
+   * If original contained only one row (good or bad), the `Parsed` must have a real
+   * `ack` action, otherwise if it has been accompanied by other rows, only the last
+   * element from the original will contain the `ack`, all others just `noop`
+   */
+  type Parsed[F[_], A, B] = Either[Payload[F, A], Payload[F, B]]
+
+  /** Apply `f` function to all elements in a list, except last one, where `lastF` applied */
+  def mapWithLast[A, B](f: A => B, lastF: A => B)(as: List[A]): Stream[Pure, B] = {
```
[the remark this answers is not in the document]
chuwy: Just cryptic without a property-based test - going to write it!

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -0,0 +1,113 @@
...
+        Sync[F].delay(reporter.report())
+
+      def enrichLatency(collectorTstamp: Option[Long]): F[Unit] =
+        collectorTstamp match {
+          case Some(delay) =>
+            Sync[F].delay {
+              registry.remove(LatencyGaugeName)
+              val _ = registry.register(LatencyGaugeName,
+                                        new Gauge[Long] {
+                                          def getValue: Long = System.currentTimeMillis() - delay
```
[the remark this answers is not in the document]
chuwy: I dunno. We can use it for testing purposes (but cannot come up with a valuable test). Otherwise - everything inside `Sync[F].delay` block supposed to be impure.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -0,0 +1,147 @@
...
+               Sync[F].unit
+           }
+    } yield Payload(List(badRow.invalid), Sync[F].unit)
+
+  /** Build a `generic_error` bad row for unhandled runtime errors */
+  def genericBadRow(
+    row: Array[Byte],
+    time: Instant,
+    error: Throwable
+  ): BadRow = {
```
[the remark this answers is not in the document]
chuwy: Will do.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -0,0 +1,147 @@
...
+    enrichLatency: Option[Long] => F[Unit]
+  )(
+    row: Payload[F, Array[Byte]]
+  ): F[Result[F]] = {
+    val payload = ThriftLoader.toCollectorPayload(row.data, processor)
+    val collectorTstamp = payload.toOption.flatMap(_.flatMap(_.context.timestamp).map(_.getMillis))
+
+    val result =
+      for {
+        _ <- Logger[F].debug(payloadToString(payload))
```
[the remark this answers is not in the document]
chuwy: Yes, I do. It's obviously for super rare situations where we can print the data being processed, but in case we'll need this again - we won't have to release a new asset for debugging only purposes.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -0,0 +1,147 @@
...
+
+    result.handleErrorWith(sendToSentry[F](row, sentry))
+  }
+
+  /** Stringify `ThriftLoader` result for debugging purposes */
+  def payloadToString(payload: ValidatedNel[BadRow.CPFormatViolation, Option[CollectorPayload]]): String =
+    payload.fold(_.asJson.noSpaces, _.map(_.toBadRowPayload.asJson.noSpaces).getOrElse("None"))
+
+  /** Transform enriched event into canonical TSV */
+  def encodeEvent(enrichedEvent: EnrichedEvent): String =
```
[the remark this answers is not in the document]
chuwy: Agreed. I actually think it should be a part of standard `Event` definition with ton of property-based tests: https://github.com/snowplow/snowplow-scala-analytics-sdk/issues/97

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
```diff
@@ -0,0 +1,289 @@
...
+/**
+ * Functions responsible for periodic assets (such as MaxMind/IAB DBs) updates
+ * The common logic is to periodically invoke a function that:
+ * 1. Downloads a file (in background) to a temp location
+ * 2. Compares file's checksum with existing one (stored in a mutable hashmap)
+ * 3. If checksums match - delete the temp file, return
+ * 4. If checksums don't match - send a signal to stop raw stream
+ * (via `SignallingRef` in [[Environment]])
+ * 5. Once raw stream is stopped - delete an old file and move
+ * temp file to the old's file location
```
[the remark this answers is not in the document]
chuwy: Yes! https://github.com/snowplow/enrich/blob/cc5d6b1df2dbdaada9e27560849db46e604f1540/modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala#L118-L172

And in fact I revealed that it is true and it does work with a handle. That's why we keep [`Enrichments`](https://github.com/snowplow/enrich/blob/cc5d6b1df2dbdaada9e27560849db46e604f1540/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala#L64) in a `Ref` (mutable variable) - that's basically why we have [reinitialize](https://github.com/snowplow/enrich/blob/cc5d6b1df2dbdaada9e27560849db46e604f1540/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala#L180) all enrichments - just to make sure that all file handles and caches are updated.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
chuwy: That's actually a very good point. Right now we do this during [`Environment` creation](https://github.com/snowplow/enrich/blob/cc5d6b1df2dbdaada9e27560849db46e604f1540/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala#L127), but maybe this is a much better place.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
chuwy: It's paused by default, when the whole thing starts https://github.com/snowplow/enrich/blob/feature/fs2-enrich/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala#L125

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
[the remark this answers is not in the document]
chuwy: I dunno. We can use it for testing purposes (but cannot come up with a valuable test). Otherwise - everything inside `Sync[F].delay` block supposed to be impure.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
chuwy: Argh. Just started to refactor it and remembered why I've put initial assets download into `Environment` initialization and not here. It's because we need to have all enrichment assets at the moment when enrichment configs are parsed and initialized, because SCE requires file to be there at the initialization point.

So we cannot do that unfortunately. I'll add a comment.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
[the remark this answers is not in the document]
chuwy: Just in case - I wrote a small benchmarking module (https://github.com/snowplow/enrich/issues/370), which should make it easy to confirm/refute performance benefiits.

chuwy:
Hey @benjben! I adressed all your feedback, added few more tests and couple of tickets (#370 - depends on NH, #371 was also discovered while I was writing tests). If anyone else from @snowplow/com-snowplowanalytics-engineering-datacapability wants to have a look - you're welcome. Otherwise this should be ready.

config/config.fs2.hocon.sample
```diff
@@ -0,0 +1,23 @@
+auth = {
+  type = "Gcp"
```
oguzhanunlu, line 2: - Acronyms are generally all-caps. Is there a specific reason to use `Gcp`?
- From a user's perspective, it'd be useful if we could see valid values of configuration fields, e.g. is `gcp` or `GCP` valid here? Or do we want to rely on user-friendly error messages explaining what's wrong and how to fix?

oguzhanunlu: - It'd be nice to see which fields are optional and and which values are used by default when applicable e.g. `assetsUpdatePeriod` if it is not configured

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
```diff
@@ -0,0 +1,296 @@
...
+   */
+  final case class State[F[_]](
+    files: Ref[F, Map[URI, Hash]],
+    pauseEnrich: Ref[F, Boolean],
+    clients: Clients[F]
+  )
+
+  object State {
+
+    /** Test pair is used in tests to initialize HTTP client */
```
oguzhanunlu, line 74: Why do we have it here if it is used in tests?

chuwy: Because we need to skip it here https://github.com/snowplow/enrich/pull/328/files/72e8d30543915d21af2150223cb1598d043e754a#diff-638a4515052fa93eecdd441517d16652eb2cd047af5d77847997c83602d1eee9R92

config/config.fs2.hocon.sample
chuwy: Spelling was dictated by codecs deriivation (it uses a exact case class name). I decided no to change it for now as `Gcp` is the only valid value here, but I agree this is something that should be fixed.

I'll add comments to the config file.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Assets.scala
oguzhanunlu: Looking at L92, it seems it isn't used in tests only, could you update this scaladoc?

oguzhanunlu: I didn't notice your comment and posted the last one, still scaladoc needs an update I think

lukeindykiewicz (approved):
Looks good in general! It's a bit to big PR to read it very carefully.

Issue #11, opened 2020-06-19
Scala Common Enrich: deprecate and remove CloudfrontLoader


chuwy:
AWS is changing its Cloudfront format.

> We are contacting you because your AWS account contains at least one CloudFront web distribution with logging enabled. Starting December 12, 2019, Amazon CloudFront is adding the following seven new fields to access logs:
> 
> • c-port – The port number of the request from the viewer.
> • time-to-first-byte – The number of seconds between receiving the request and writing the first byte of the response, as measured on the server.
> • x-edge-detailed-result-type – When the result type is an error, this field contains the specific type of error.
> • sc-content-type – The value of the HTTP Content-Type header of the response.
> • sc-content-len – The value of the HTTP Content-Length header of the response.
> • sc-range-start – When the response contains the HTTP Content-Range header, this field contains the range start value.
> • sc-range-end – When the response contains the HTTP Content-Range header, this field contains the range end value.

benjben:
Remove `Absence of CollectorPayload (None) should be supported                                   $e4` in `EtlPipelineSpec.scala` when this happens

PR #336, opened 2020-09-08
Fix PII pseudonymizer enrichment + unit tests

benjben:
The test failing is due to http://snowplow-hosted-assets.s3.amazonaws.com/third-party/maxmind/GeoLite2-City.mmdb being removed/empty

This will be fixed in another commit.


modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/EtlPipelineSpec.scala
```diff
@@ -44,6 +45,8 @@ class EtlPipelineSpec extends Specification with ValidatedMatchers {
   def is = s2"""
   EtlPipeline should always produce either bad or good row for each event of the payload   $e1
   Processing of events with malformed query string should be supported                     $e2
+  Processing of invalid CollectorPayload (CPFormatViolation bad row) should be supported   $e3
+  Absence of CollectorPayload (None) should be supported                                   $e4
```
chuwy, line 49: IIRC we need this because Cloudfront collector supported the format where the first line of file was starting with a comment, like:

```
# timestamp ip ...
2020-02-01 01:32:43 126.10.1.2 ...
```

Need to decide when we're going to drop that officially. Also probably worth to mention it in the test: https://github.com/snowplow/enrich/issues/11

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/EnrichmentManagerSpec.scala
```diff
@@ -345,6 +345,21 @@ class EnrichmentManagerSpec extends Specification with EitherMatchers {
       EnrichmentManager.getIabContext(input, iabEnrichment) must beRight.like { case ctx => ctx must beSome }
     }
   }
+
+  "getCollectorVersionSet" should {
+    "return an enrichment failure if v_collector is null or empty" >> {
+      val input = new EnrichedEvent()
+      EnrichmentManager.getCollectorVersionSet(input) must beLeft.like { case _: FailureDetails.EnrichmentFailure => ok }
+      input.v_collector = ""
+      EnrichmentManager.getCollectorVersionSet(input) must beLeft.like { case _: FailureDetails.EnrichmentFailure => ok }
```
chuwy, line 354: Also worth to add `ko` branch.

chuwy (approved):
Looks great!

dilyand:
> @dilyand don't we want to leave an empty string empty in the hashed output ?

@benjben I'm not sure. But:
- If we leave an empty string empty but hash all other strings, we are making a decision that an empty string is not PII and so does not need to be pseudonymized.
- Depending on context, we might make that call in other, more complex situations. For example, if the field is `email_address` but the value does not match a 'valid email' regex, then it's not PII and we can leave it unhashed.
- Fundamentally, both of the above are the same type of sanitisation, the only difference is the complexity of the 'regex' that needs to be matched.
- Since we're not doing 2, we shouldn't be doing 1.
- Maybe in some situations knowing that a field is an empty string will reveal some PII? (Though this point is weakened by the fact that all empty strings will have the same hash, so they will be easy to recognise.)
- The case with `null` value is different. There we return no value and not a hash, because there is nothing to hash, not because we decide it's not sensitive.

I'm happy to consider counterpoints, but with the above in mind I'm leaning towards hashing the empty string.

WDYT?

benjben:
> If we leave an empty string empty but hash all other strings, we are making a decision that an empty string is not PII
IMHO that's the case. And hashing the empty strings will always produce the same hash.

> if the field is email_address but the value does not match a 'valid email' regex, then it's not PII and we can leave it unhashed.
IMHO hashing of PII is not related to data validation and we should hash whatever there is to hash.

dilyand:
In that case, the test reveals a bug...

I think I agree with: 

> hashing of PII is not related to data validation and we should hash whatever there is to hash

but we differ in that I think this includes all values, including empty string.

Btw, do you want to keep this test? I can rebase to merge my commit with your first one.

benjben:
> In that case, the test reveals a bug...

But that doesn't reproduce the error `""` => object :thinking: 

> do you want to keep this test?

Sure !! The more testing the better !

chuwy:
@dilyand, @benjben let me know when its ready for review again - I'd like to start putting 1.4.0 together.

chuwy:
Hey @benjben! Are you ok to fix the #351 in this branch and potentially incorporate Dilyan's commit.

benjben:
Sure @chuwy ! So just to be sure that we don't duplicate work, this is what you first wanted to take care of but finally I do right ?

chuwy:
@benjben nope, or at least I don't remember taking care of it. I implemented the  recovery for a client.

chuwy:
Oh, actually you're right - I wanted to do this! Sorry. Yeah, let's decide tomorrow - I'm just signing off today and thought you'll be working on something. Ignore me.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
```diff
@@ -132,6 +133,37 @@ object PiiPseudonymizerEnrichment extends ParseableEnrichment {
...
+              case Some(orig_arr) =>
+                // we can use zip because there should never be new fields in an array, only in an object
+                val values = hashed_arr.zip(orig_arr).map { case (hashed, orig) => removeAddedFields(hashed, orig) }
+                Json.fromValues(values)
+              case None =>
+                Json.fromValues(hashed_arr) // should never happen. Would mean change of type of one field
+            }
+          case None => hashed // hashed is neither JSON object nor array
+        }
+    }
```
chuwy, line 166: My modest implementation:

```suggestion
  private[pii] def removeAddedFields(hashed: Json, original: Json): Json = {
    val fixedObject = for {
      hashedFields <- hashed.as[Map[String, Json]]
      originalFields <- original.as[Map[String, Json]]
      newMap = hashedFields.flatMap {
        case (k, v) => originalFields.get(k) match {
          case Some(originalValue) => Some((k, removeAddedFields(v, originalValue)))
          case None => None
        }
      }
    } yield Json.fromFields(newMap)

    lazy val fixedArray = for {
      hashedArr <- hashed.asArray
      originalArr <- original.asArray
      newArr = hashedArr.zip(originalArr).map { case (hashed, orig) => removeAddedFields(hashed, orig) }
    } yield Json.fromValues(newArr)

    fixedObject.toOption.orElse(fixedArray).getOrElse(hashed)
  }
```

chuwy (approved):
Not sure I understand evertying that is going on here, but happy that we have tests.

My implementation of `removeAddedFields`  is here just for a reference, feel free to add or ignore, although it's a bit shorter.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
chuwy: I'd personally replace `flatMap` on 142 with `unite` as I hate `A => Option[B]` flatMapping on `List[A]`, but it will require additional cats import.

oguzhanunlu (approved):
LGTM!

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/pii/PiiPseudonymizerEnrichment.scala
benjben: Thanks for the better implementation @chuwy !

Issue #269, opened 2020-06-26
Beam: use test Maxmind databases


chuwy:
Whenever a new version of the Maxmind DB for ip lookups is uploaded to the hosted assets bucket, the `EnrichWithLocalFileSpec` test in Beam Enrich can start failing if the data points associated with the IP address used in it have changed.

Maxmind provides stable test DBs that we can use to make the test less flaky: https://github.com/maxmind/MaxMind-DB/tree/master/test-data

Moved from: https://github.com/snowplow/snowplow/issues/4242

chuwy:
Migrated from https://github.com/snowplow/beam-enrich/issues/22 (comments are auto-generated)

PR #337, opened 2020-09-08
Use test Maxmind databases

oguzhanunlu:
This PR bundles a test maxmind db with the test suite of beam. For visibility, the db in beam module contains a single record as following -- copy pasting from my perl script

```perl
'37.228.225.32/32' => {
        country => {
            confidence => 95,
            geoname_id => 2635167,
            is_in_european_union => 'true',
            iso_code => 'IE',
            names => {
                en => 'Ireland'
            }
        },
        city => {
            confidence => 80,
            geoname_id => 2655045,
            names => {
                en => 'Dublin'
            }
        },
        subdivisions => [
            {
                confidence => 70,
                geoname_id => 6269131,
                iso_code   => 'L',
                names => {
                    en => 'Leinster'
                }
            },
        ],
        postal => {
            code => 'D02',
            confidence => 100
        },
        location => {
            accuracy_radius => 100,
            latitude => 53.3331,
            longitude => -6.2489
        }
   }
```

I created the database using Maxmind's tooling but it'd be nice to document the know-how somewhere, since it is likely we'll want to modify existing record or add records etc. Where would be suitable for this documentation? @benjben @chuwy


chuwy:
> Where would be suitable for this documentation?

That's a good question. How about documenting it in the issue (https://github.com/snowplow/enrich/issues/269) and add

```
// In order to create new DB see https://github.com/snowplow/enrich/issues/269
```

In the spec to not forget about it. WDYT?

oguzhanunlu:
That sounds good to me @chuwy , will do so soon.

benjben:
Shoudn't we be hosting the db on GCS rather than Github given that it's binary?

chuwy:
> Shoudn't we be hosting the db on GCS rather than Github given that it's binary?

Also a good point. I think maybe its okay given that size is just 1KB and that we already store couple other binary assets, but I won't oppose hosting it on GCS.

chuwy:
Hey @oguzhanunlu! Could you please do the same for Stream and Common Enrich and submit the PR against https://github.com/snowplow/enrich/pull/326. I'd like to merge it to `develop`, but cannot since tests are failing.

oguzhanunlu:
sure @chuwy !

oguzhanunlu:
@chuwy , I changed the base branch to `feature/cats2`, rebased my branch on top of that and added remaining commits

chuwy:
Actually, it's passed all tests - no need to rebase to `feature/fs2-enrich`. Changed branch back to `develop` and going to merge. @benjben could you please also review it?

chuwy (approved):

benjben (commented):
The binary files `GeoIP2-City.mmdb` and `GeoIP2-ISP.mmdb` in the test resources of `common/` are not mentionned in `IpLookupsEnrichmentSpec` in `common/`, are they used?

Otherwise still not a fan of having multiple binary 70kb files on Github.

oguzhanunlu:
Hey @benjben , thanks for the review! Let me answer your questions.

> The binary files ... are they used?

Yes, see [L56](https://github.com/snowplow/enrich/blob/master/modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/IpLookupsEnrichmentSpec.scala#L56) where `localMode` is `true` which means the db files specified in the enrichment config should be fetched from local resources.

> Otherwise still not a fan of having multiple binary 70kb files on Github.

This is not the case in this PR. In fact, if you check master, there are db files with size [18.9 KB](https://github.com/snowplow/enrich/blob/master/modules/common/src/test/resources/com/snowplowanalytics/snowplow/enrich/common/enrichments/registry/GeoIP2-City.mmdb) and [73.4 KB](https://github.com/snowplow/enrich/blob/master/modules/common/src/test/resources/com/snowplowanalytics/snowplow/enrich/common/enrichments/registry/GeoIP2-ISP.mmdb). This PR reduces their size to [1.43 KB](https://github.com/snowplow/enrich/blob/add-test-maxmind-db/modules/common/src/test/resources/com/snowplowanalytics/snowplow/enrich/common/enrichments/registry/GeoIP2-City.mmdb) and [2.45 KB](https://github.com/snowplow/enrich/blob/add-test-maxmind-db/modules/common/src/test/resources/com/snowplowanalytics/snowplow/enrich/common/enrichments/registry/GeoIP2-ISP.mmdb) respectively.

benjben (approved):
Thanks @oguzhanunlu !

PR #342, opened 2020-09-10
Common: handle empty query string parameters in adapters

benjben:


oguzhanunlu (approved):
Thanks @benjben !

chuwy (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/snowplow/RedirectAdapter.scala
```diff
@@ -96,6 +96,16 @@ object RedirectAdapter extends Adapter {
               msg
             )
           Monad[F].pure(failure.invalidNel)
+        case Some(null) =>
```
chuwy, line 99: Actually, I think:

1. We should not distinguish between `None` and `Some(null)`.
2. We need to check this pattern `originalParams.get(x)` in all other adapters. Can you do that, @benjben?

benjben: > We should not distinguish between None and Some(null)

`None` is for when `u` param is absent, `Some(null)` is for `u` exists but doesn't have a value (`&u`). Don't we want to distinguish the 2 cases in the error message to help to fix the tracking?

> Can you do that, @benjben?

Sure

chuwy: > Don't we want to distinguish the 2 cases in the error message to help to fix the tracking?

In that case, we at least should change `toMap` signature to `Map[String, Option[String]]`, but I personally don't think that it such a big deal to be specific - it should be easy to debug with payload anyway. Just `Some(null)` is such a horrible code smell.

benjben: > it should be easy to debug with payload anyway

You're right we have the raw payload with the raw query string anyway. I'll remove this `Some(null)`

benjben:
I updated [toMap](https://github.com/snowplow/enrich/pull/342/files#diff-cacb91ef1cf87126529fd8190ff84c4aR129-R130) return `Map[String, Option[String]]`.

To help the refactoring of unit tests I created [this helper](https://github.com/snowplow/enrich/pull/342/files#diff-88b94402054be7151117ddb1a87c1fd6R108).

chuwy (approved):
Looks great!

I personally think it would be better to squash commits under single ticket, like `Common: handle empty query parameters`, but I'm okay if it will be #342 as well.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/SpecHelpers.scala
```diff
@@ -103,4 +103,8 @@ object SpecHelpers {
       .leftMap(err => s"Can't parse [$rawJson] as Json, error: [$err]")
       .flatMap(SelfDescribingData.parse[Json])
       .leftMap(err => s"Can't parse Json [$rawJson] as as SelfDescribingData, error: [$err]")
+
+  implicit class MapOps[A, B](underlying: Map[A, B]) {
+    def toOpt: Map[A, Option[B]] = underlying.map { case (a, b) => (a, Option(b)) }
```
oguzhanunlu, line 108: Nice trick!

oguzhanunlu (approved):

Issue #374, opened 2020-10-12
Common: fix API Request Enrichment output deserialization


oguzhanunlu:
`io.circe.jackson.mapper` deserializes all integers and numbers to `JsonBigDecimal` and downstream in the codebase, it is converted to `DecimalNode` by `circeToJackson()` in `iglu-scala-client`, and the validator treats `DecimalNode` as `number` which results in a bad row with `number found, integer expected` error message in case output schema of API Request Enrichment contains `integer` property

We'll introduce a custom mapper whose only difference compared to the original mapper is how integers are deserialized. Unlike the original, it'll deserialize integers to `JsonLong`.

Note that json schemas might contain integer values that don't fit into `Long` and this case will also result in a bad row, but we don't expect to see this case in real life.

The actual [diff](https://github.com/snowplow/enrich/pull/375/commits/606e074e0b971f788bf709573bb9fdf50a2ae1d0#diff-52c2ad6b30f0646aeac2fc5bf250fd6ed3a4ffc5e9c3accea22f9557b7b6e9dfL35-R33) happens in `CirceJsonDeserializer.deserialize` where

```scala
case JsonTokenId.ID_NUMBER_INT | JsonTokenId.ID_NUMBER_FLOAT =>
  (Some(Json.JNumber(JsonBigDecimal(jp.getDecimalValue))), parserContext)
```

is replaced by

```scala
case JsonTokenId.ID_NUMBER_INT => (Some(Json.JNumber(JsonLong(jp.getLongValue))), parserContext)
case JsonTokenId.ID_NUMBER_FLOAT => (Some(Json.JNumber(JsonBigDecimal(jp.getDecimalValue))), parserContext)
```

chuwy:
Bug manifests itself as an invaildation of a schema, added by API Request enrichment. If output schema requires integer type, the validator complains that "42 is a number, but integer expected", because in its internal representation, 42 is a number.

chuwy:
What about SQL Query enrichment, @oguzhanunlu? It's inputs also seem to be affected, but I don't see any changes regarding to that.

PR #375, opened 2020-10-12
Common: fix API Request Enrichment output deserialization

oguzhanunlu:
This PR introduces a new package `io.circe.jackson.enrich` which contains a custom deserialization logic since `circe-jackson`'s deserializer converts a `Long` to `JsonBigDecimal`, which results in `number found, integer expected` enrichment failures.

TO DO

- [ ] Check SQL query enrichment https://github.com/snowplow/enrich/issues/374#issuecomment-707110319

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/CirceUtils.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/CirceUtils.scala
@@ -15,8 +15,9 @@ package utils
 
 import cats.data.Validated
 import cats.syntax.either._
-
+import com.fasterxml.jackson.databind.ObjectMapper
 import io.circe._
+import io.circe.jackson.enrich.CirceJsonModule
 
 object CirceUtils {
 
@@ -43,4 +44,19 @@ object CirceUtils {
         s"Could not extract $pathStr as $clas from supplied JSON due to ${e.getMessage}"
       }
   }
+
+  /**
+   * A custom ObjectMapper specific to Circe JSON AST
+   *
+   * The only difference from the original mapper `io.circe.jackson.mapper` is
+   * how `Long` is deserialized. The original mapper maps a `Long` to `JsonBigDecimal`
+   * whereas this custom mapper deserializes a `Long` to `JsonLong`.
+   *
+   * This customization saves Snowplow events from failing when derived contexts are
+   * validated post-enrichment. If output schema of API Request Enrichment has an integer
+   * field, `JsonBigDecimal` representation of a Long results in a bad row
+   * with message `number found, integer expected` in Iglu Scala Client, since jackson
+   * treats `DecimalNode` as number in all cases.
+   */
+  final val mapper: ObjectMapper = (new ObjectMapper).registerModule(CirceJsonModule)
 }
--- a/modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
+++ b/modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
@@ -1,0 +1,89 @@
+package io.circe.jackson.enrich
+
+import java.util
+
+import com.fasterxml.jackson.core.{JsonParser, JsonTokenId}
+import com.fasterxml.jackson.databind.{DeserializationContext, JsonDeserializer}
+import io.circe.jackson.{DeserializerContext, JacksonCompat, ReadingList, ReadingMap}
+import io.circe.{Json, JsonBigDecimal, JsonLong}
+
+import scala.annotation.{switch, tailrec}
+import scala.collection.JavaConverters._
+
+private[jackson] final class CirceJsonDeserializer(klass: Class[_]) extends JsonDeserializer[Object] with JacksonCompat {
+  override def isCachable: Boolean = true
+
+  override def deserialize(jp: JsonParser, ctxt: DeserializationContext): Json = {
+    val value = deserialize(jp, ctxt, List())
+    if (!klass.isAssignableFrom(value.getClass)) handleUnexpectedToken(ctxt)(klass, jp)
+
+    value
+  }
+
+  @tailrec
+  def deserialize(
+    jp: JsonParser,
+    ctxt: DeserializationContext,
+    parserContext: List[DeserializerContext]
+  ): Json = {
+    if (jp.getCurrentToken == null) jp.nextToken()
+
+    val (maybeValue, nextContext) = (jp.getCurrentToken.id(): @switch) match {
+      case JsonTokenId.ID_NUMBER_INT => (Some(Json.JNumber(JsonLong(jp.getLongValue))), parserContext)
+      case JsonTokenId.ID_NUMBER_FLOAT => (Some(Json.JNumber(JsonBigDecimal(jp.getDecimalValue))), parserContext)
+      case JsonTokenId.ID_STRING => (Some(Json.JString(jp.getText)), parserContext)
+      case JsonTokenId.ID_TRUE => (Some(Json.JBoolean(true)), parserContext)
+      case JsonTokenId.ID_FALSE => (Some(Json.JBoolean(false)), parserContext)
+      case JsonTokenId.ID_NULL => (Some(Json.JNull), parserContext)
+      case JsonTokenId.ID_START_ARRAY => (None, ReadingList(new util.ArrayList) +: parserContext)
+
+      case JsonTokenId.ID_END_ARRAY =>
+        parserContext match {
+          case ReadingList(content) :: stack =>
+            (Some(Json.fromValues(content.asScala)), stack)
+          case _ => throw new IllegalStateException("Jackson read ']' but parser context is not an array")
+        }
+
+      case JsonTokenId.ID_START_OBJECT => (None, ReadingMap(new util.ArrayList) +: parserContext)
+
+      case JsonTokenId.ID_FIELD_NAME =>
+        parserContext match {
+          case (c: ReadingMap) :: stack => (None, c.setField(jp.getCurrentName) +: stack)
+          case _ =>
+            throw new IllegalStateException("Jackson read a String field name but parser context is not a json object")
+        }
+
+      case JsonTokenId.ID_END_OBJECT =>
+        parserContext match {
+          case ReadingMap(content) :: stack =>
+            (
+              Some(Json.fromFields(content.asScala)),
+              stack
+            )
+          case _ => throw new IllegalStateException("Jackson read '}' but parser context is not a json object")
+        }
+
+      case JsonTokenId.ID_NOT_AVAILABLE =>
+        throw new IllegalStateException("Jackson can't return the json token yet")
+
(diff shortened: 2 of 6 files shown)


modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
```diff
@@ -0,0 +1,101 @@
...
+ * and you may not use this file except in compliance with the Apache License Version 2.0.
+ * You may obtain a copy of the Apache License Version 2.0 at http://www.apache.org/licenses/LICENSE-2.0.
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+
+package io.circe
```
chuwy, line 14: Could we please get rid of these nested packages, unless they're necessary. They make imports look very confusing.

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
```diff
@@ -0,0 +1,101 @@
...
+        }
+
+      case JsonTokenId.ID_END_OBJECT =>
+        parserContext match {
+          case ReadingMap(content) :: stack =>
+            (
+              Some(Json.fromFields(content.asScala)),
+              stack
+            )
+          case _ => throw new RuntimeException("We weren't reading an object, something went wrong")
```
chuwy, line 75: "We" sounds very confusing here. Is it an original message? Why isn't it `InvalidArgumentException` or `InvalildStateException`? `RuntimeException` is very unspecific.

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
```diff
@@ -0,0 +1,101 @@
+/*
+ * Copyright (c) 2012-2020 Snowplow Analytics Ltd. All rights reserved.
```
chuwy, line 2: I don't think we can use this copyright.

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonModule.scala
```diff
@@ -0,0 +1,60 @@
...
+package jackson
+package enrich
+
+import com.fasterxml.jackson.core.Version
+import com.fasterxml.jackson.databind.deser.Deserializers
+import com.fasterxml.jackson.databind.module.SimpleModule
+import com.fasterxml.jackson.databind.ser.Serializers
+import com.fasterxml.jackson.databind._
+
+object CirceJsonModule extends SimpleModule("CirceJson", Version.unknownVersion()) {
```
chuwy, line 24: `CirceJson` seems to be a global name, so we can have a clash.

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonModule.scala
```diff
@@ -0,0 +1,60 @@
...
+    context.addDeserializers(
+      new Deserializers.Base {
+        override final def findBeanDeserializer(
+          javaType: JavaType,
+          config: DeserializationConfig,
+          beanDesc: BeanDescription
+        ) = {
+          val klass = javaType.getRawClass
+          if (classOf[Json].isAssignableFrom(klass) || klass == Json.JNull.getClass)
+            new io.circe.jackson.enrich.CirceJsonDeserializer(klass)
```
chuwy, line 36: Weird class path.

modules/common/src/main/scala/io/circe/jackson/enrich/package.scala
```diff
@@ -0,0 +1,21 @@
...
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+
+package io.circe
+package jackson
+
+import com.fasterxml.jackson.databind.ObjectMapper
+
+package object enrich {
+  val mapper: ObjectMapper = (new ObjectMapper).registerModule(io.circe.jackson.enrich.CirceJsonModule)
```
chuwy, line 20: Don't see a reason why it should be in `io.circe` package.

chuwy (changes requested):
Apart from things I mentioned:

1. I'd ilke to look at the diff. It seems that it should be a fairly small change compared to an original implementation, just taking care of `Long`s and `Decimal`s.
2. it needs **a lot** of tests. A don't see any.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/CirceUtils.scala
```diff
@@ -43,4 +44,6 @@ object CirceUtils {
         s"Could not extract $pathStr as $clas from supplied JSON due to ${e.getMessage}"
       }
   }
+
+  final val mapper: ObjectMapper = (new ObjectMapper).registerModule(CirceJsonModule)
```
chuwy, line 48: Worth noting why this mapper exists.

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
```diff
@@ -0,0 +1,89 @@
...
+      case JsonTokenId.ID_TRUE => (Some(Json.JBoolean(true)), parserContext)
+      case JsonTokenId.ID_FALSE => (Some(Json.JBoolean(false)), parserContext)
+      case JsonTokenId.ID_NULL => (Some(Json.JNull), parserContext)
+      case JsonTokenId.ID_START_ARRAY => (None, ReadingList(new util.ArrayList) +: parserContext)
+
+      case JsonTokenId.ID_END_ARRAY =>
+        parserContext match {
+          case ReadingList(content) :: stack =>
+            (Some(Json.fromValues(content.asScala)), stack)
+          case _ => throw new RuntimeException("Jackson read ']' but parser context is not an array")
```
chuwy, line 44: Still plain `RuntimeException`

modules/common/src/main/scala/io/circe/jackson/enrich/CirceJsonDeserializer.scala
```diff
@@ -0,0 +1,89 @@
+package io.circe.jackson
```
chuwy, line 1: Still a nested package.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/JsonPathSpec.scala
```diff
@@ -88,4 +93,23 @@ class JsonPathSpec extends Specification {
 
   def e6 =
     JsonPath.query("$.store.book[2]", Json.fromString("somestring")) must beRight(List())
+
+  def e7 =
+    JsonPath.anyToJson(null) must be(Json.Null)
+
+  def e8 = {
+    JsonPath.anyToJson(5) must beEqualTo(Json.fromLong(5))
```
chuwy, line 101: This one won't have an effect - you can change it to `JsonPath.anyToJson(5) must beEqualTo(Json.fromLong(10))`

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/JsonPathSpec.scala
```diff
@@ -88,4 +93,23 @@ class JsonPathSpec extends Specification {
...
+    JsonPath.anyToJson(-1) must beEqualTo(Json.fromLong(-1))
+  }
+
+  def e9 = {
+    JsonPath.anyToJson(3.99) must beEqualTo(Json.fromBigDecimal(3.99))
```
chuwy, line 106: This one won't have an effect.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/JsonPathSpec.scala
chuwy: Implicit widening.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/JsonPathSpec.scala
```diff
@@ -88,4 +93,23 @@ class JsonPathSpec extends Specification {
 
   def e6 =
     JsonPath.query("$.store.book[2]", Json.fromString("somestring")) must beRight(List())
+
+  def e7 =
+    JsonPath.anyToJson(null) must be(Json.Null)
+
+  def e8 = {
+    JsonPath.anyToJson(5) must beEqualTo(Json.fromLong(5))
+    JsonPath.anyToJson(-1) must beEqualTo(Json.fromLong(-1))
```
chuwy, line 102: Am I right that this test would fail with original mapper?

chuwy (changes requested):
Tests for `Short`, `Float`,  `Long` (in `e5` you use implicit widening) are missing.

Also need to add something more complicated:

```scala
JsonPath.query("$.store[2]", json"""{"store": [null,1,2]}""") must beRight(List(something))
```

Also, as far as I understand from [Ben's ticket](https://github.com/circe/circe-jackson/issues/49) `Json`-based constructors were creating different instances compared to `parse` and literal JSON. How do we test those cases?

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/JsonPathSpec.scala
oguzhanunlu: I changed the value to a long in my last commit but it doesn't fail using original mapper, since equality check doesn't include type check but value check only. However ValidatorSpec added in last commit fails using original mapper.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/apirequest/ValidatorSpec.scala
```diff
@@ -0,0 +1,67 @@
...
+    validate number field using a positive double value                          $e6
+  """
+
+  val schema =
+    json"""{ "type": "object", "properties": { "orderID": { "type": "integer" }, "price": { "type": "number" } }, "additionalProperties": false }"""
+
+  def e1 =
+    query("$", json"""{"orderID": 9223372036854775807 }""")
+      .fold(_ => false, fb => CirceValidator.validate(fb.head, schema).isRight)
+      .must(beTrue)
```
chuwy, line 40:
```suggestion
    query("$", json"""{"orderID": 9223372036854775807 }""")
      .map(fb => CirceValidator.validate(fb.head, schema)) must beRight
```

chuwy (approved):
👍 

Please squash and merge into `develop`.

PR #382, opened 2020-10-27
Release/1.4.1

oguzhanunlu:


oguzhanunlu:
I published rc1 and smoke tested successfully

.github/workflows/test.yml
```diff
@@ -24,6 +24,8 @@ jobs:
       uses: actions/setup-java@v1
       with:
         java-version: 11
+    - name: Prepare Mock server for SCE ApiRequestEnrichmentIntegrationTest (launch in background)
```
chuwy, line 27: Why do we need second one?

integration-tests/api-lookup-test.py
```diff
@@ -72,7 +72,7 @@ def generate_response(self, method, auth=None):
                 }
             }
         else:
-            response = { "serviceName": "sp-api-request-enrichment" }
```
chuwy, line 75: Why did we remove the string?

modules/common/src/main/scala/io/circe/jackson/enrich/JacksonParser.scala
```diff
@@ -0,0 +1,40 @@
+package io.circe.jackson.enrich
```
chuwy, line 1: Is it a copy from circe jackson? It feels that in this case we can actually put our copyright as you're just impelmenting the interace?

modules/common/src/main/scala/io/circe/jackson/enrich/JacksonParser.scala
```diff
@@ -0,0 +1,40 @@
...
+import scala.util.control.NonFatal
+
+trait JacksonParser extends Parser { this: WithJacksonMapper =>
+  final def parse(input: String): Either[ParsingFailure, Json] =
+    try Right(mapper.readValue(jsonStringParser(input), classOf[Json]))
+    catch {
+      case NonFatal(error) => Left(ParsingFailure(error.getMessage, error))
+    }
+
+  final def parseFile(file: File): Either[ParsingFailure, Json] =
```
chuwy, line 17: We cannot get away without all these functions?

modules/common/src/main/scala/io/circe/jackson/enrich/WithJacksonMapper.scala
```diff
@@ -0,0 +1,18 @@
+package io.circe.jackson.enrich
+
+import java.io.{File, Writer}
+
+import com.fasterxml.jackson.core.{JsonFactory, JsonParser}
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.snowplowanalytics.snowplow.enrich.common.utils.CirceUtils
+
+class WithJacksonMapper {
+  final val mapper: ObjectMapper = CirceUtils.mapper
```
chuwy, line 10: We need to note somewhere that this whole PR, all new modules are just to make this `mapper` default. Right now it looks like a huge copy-pasta if we don't point to this mapper.

modules/common/src/main/scala/io/circe/jackson/enrich/package.scala
```diff
@@ -0,0 +1,3 @@
+package io.circe.jackson
+
+package object enrich extends WithJacksonMapper with JacksonParser with JacksonCompat {}
```
chuwy, line 3: Can be `com.snowplowanalytics.snowplow.enrich.common.utils.CirceUtils.parser`, again just to make it clear why we need this.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/apirequest/ApiRequestEnrichmentIntegrationTest.scala
```diff
@@ -49,30 +49,17 @@ class ApiRequestEnrichmentIntegrationTest extends Specification {
...
-            "pojo": {
-              "field": "app_id"
-            }
-          }
-        ],
+        "inputs": [],
         "api": {
           "http": {
             "method": "GET",
-            "uri": "http://localhost:8001/guest/api/{{client}}/{{user}}?format=json",
```
chuwy, line 69: Why did you remove other important test cases for placeholders? We should be adding tests with every PR, not removing them.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/Output.scala
```diff
@@ -117,6 +116,5 @@ final case class JsonOutput(jsonPath: String) extends ApiOutput[Json] {
       case other => other.leftMap(JsonPathException.apply)
     }
 
-  def parseResponse(response: String): Either[Throwable, Json] =
-    parse(response)
+  def parseResponse(response: String): Either[Throwable, Json] = io.circe.jackson.enrich.parse(response)
```
chuwy, line 119: It feels that all we need here is
```suggestion
  private[this] final val jsonFactory: JsonFactory = new JsonFactory(CirceUtils.mapper)
  def parseResponse(response: String): Either[Throwable, Json] = {
    val parser = jsonFactory.createParser(input)
    CirceUtils.mapper.readValue(parser(input), classOf[Json])
  }
```

And we don't need the rest of the magic with creating own parsers.

chuwy (changes requested):
Quite a few things to improve:

1. Tests got worse. We removed important cases and didn't add anything new. Also, a lot of stuff can be tested with unit tests and we haven't write any.
2. Feels like a lot of unnecessary and obscure code duplication.
3. But most important is that I don't get the mechanics of the problem. Jawn should be parsing numbers just fine, I assume into `JsonNumber`? Likely, the problem happens at the moment where we transform `Json` into `JsonNode`. And we have full control over it (?). So why cannot we fix it there?

integration-tests/api-lookup-test.py
oguzhanunlu: It wasn't being used

.github/workflows/test.yml
oguzhanunlu: yeah, I can merge both usages

modules/common/src/main/scala/io/circe/jackson/enrich/JacksonParser.scala
oguzhanunlu: yes, it is copy

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/apirequest/ApiRequestEnrichmentIntegrationTest.scala
oguzhanunlu: placeholders weren't being used by the api, so I removed them

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/Output.scala
```diff
@@ -117,6 +117,9 @@ final case class JsonOutput(jsonPath: String) extends ApiOutput[Json] {
       case other => other.leftMap(JsonPathException.apply)
     }
 
+  private[this] val jsonFactory: JsonFactory = new JsonFactory(CirceUtils.mapper)
   def parseResponse(response: String): Either[Throwable, Json] =
-    parse(response)
+    Either
+      .catchNonFatal[JsonParser](jsonFactory.createParser(response))
+      .flatMap(parser => Either.catchNonFatal[Json](CirceUtils.mapper.readValue(parser, classOf[Json])))
```
chuwy, line 124:
```suggestion
    Either
      .catchNonFatal[Json] { 
        val parser = jsonFactory.createParser(response)
        CirceUtils.mapper.readValue(parser, classOf[Json])
      }
```

No reasons to compose these - it's just unnecessary boxing.

chuwy (commented):
Looks much better.

Waiting for more tests.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/Output.scala
```diff
@@ -118,5 +116,5 @@ final case class JsonOutput(jsonPath: String) extends ApiOutput[Json] {
     }
 
   def parseResponse(response: String): Either[Throwable, Json] =
-    parse(response)
+    io.circe.jackson.enrich.parse(response)
```
chuwy, line 119: Why not to import it? Everything else is not fully-qualified.

modules/common/src/main/scala/io/circe/jackson/enrich/package.scala
```diff
@@ -0,0 +1,20 @@
+package io.circe.jackson
+
+import com.fasterxml.jackson.core.{JsonFactory, JsonParser}
+import com.snowplowanalytics.snowplow.enrich.common.utils.CirceUtils
+import io.circe.{Json, ParsingFailure}
+
+import scala.util.control.NonFatal
+
+package object enrich extends JacksonCompat {
```
chuwy, line 9: 1. Explanation is missing on why we need this hack.
2. This is not a copy-paste from other lib - we should be safe putting our copyright here
3. `io.circe.jackson.enrich` is slightly confusing. `io.circe.jackson.snowplow` maybe? Not much better, but out of ideas. This one is optional.

chuwy (changes requested):
Few minor things. Also, can you please get rid of those imports shufflings? And still can write more tests I belive for similar use cases.

chuwy (approved):

Issue #388, opened 2020-10-30
Crash enrich at startup if Sentry DSN is not correct


benjben:
At the moment enrich fails only when an event is to be sent to Sentry, e.g. : 
```
[...]

PR #392, opened 2020-11-04
Beam: handle commas in Sentry DSN parameter (close #391)

### LINKED ISSUE #391: Beam: handle commas in Sentry DSN parameter
ATM if `--sentry-dsn=` parameter contains a `,`, which is needed for tags, Beam Enrich crashes with :
```
java.lang.IllegalArgumentException: Multiple values for property 'sentry-dsn'
     at com.spotify.scio.Args.optional(Args.scala:106)
     at com.snowplowanalytics.snowplow.enrich.beam.config$EnrichConfig$.$anonfun$apply$4(config.scala:80)
     at scala.util.Either.map(Either.scala:353)
     at com.snowplowanalytics.snowplow.enrich.beam.config$EnrichConfig$.$anonfun$apply$1(config.scala:62)
     at scala.util.Either.flatMap(Either.scala:341)
     at com.snowplowanalytics.snowplow.enrich.beam.config$EnrichConfig$.apply(config.scala:61)
     at com.snowplowanalytics.snowplow.enrich.beam.ConfigSpec.$anonfun$new$12(ConfigSpec.scala:140)
     at org.scalatest.OutcomeOf.outcomeOf(OutcomeOf.scala:85)
     at org.scalatest.OutcomeOf.outcomeOf$(OutcomeOf.scala:83)
     at org.scalatest.OutcomeOf$.outcomeOf(OutcomeOf.scala:104)
```
due to [args.optional()](https://github.com/spotify/scio/blob/master/scio-core/src/main/scala/com/spotify/scio/Args.scala#L108) on an argument which is a `List()`.

benjben:
Logic is:
- if `--sentry-dsn=foo` then `foo` is put in `List(foo)`
- if `--sentry-dsn=a, b,c` then `a` `b` and `c` are put in `List(a, b,c)`
So by calling `args.asMap.get("sentry-dsn").map(_.mkString(","))` we are sure to reconstruct it back and we don't need base64 encoding/decoding


modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/config.scala
```diff
@@ -78,7 +78,7 @@ object config {
         resolver,
         args.optional("enrichments"),
         args.optional("labels"),
-        args.optional("sentry-dsn"),
+        args.asMap.get("sentry-dsn").map(_.mkString(",")),
```
chuwy, line 81: Hacky, but works! Maybe worth leaving a comment that we had to do it in order to support tags?

chuwy (approved):
Looks great!

oguzhanunlu (approved):

oguzhanunlu:
Thanks @benjben !

benjben:
Hi @chuwy @oguzhanunlu , I added the implementation of #388.
While I was at it, I refactored/cleaned configuration parsing in Beam Enrich, I got rid of `ParsedEnrichConfig` in favor of validating the configuration directly when creating `EnrichConfig`. I also add unit tests to make sure that the app crashes if a config is not valid.

modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/config.scala
```diff
@@ -197,4 +200,15 @@ object config {
   def parseLabels(input: String): Either[String, Map[String, String]] =
     decode[Map[String, String]](input)
       .leftMap(_ => s"Invalid `labels` format, expected json object, received: $input")
+
+  /** Create a java.net.URI from string with Sentry DSN. */
+  private def parseSentryDsn(maybeDSN: Option[String]): Either[String, Option[URI]] =
+    maybeDSN match {
+      case Some(uri) =>
+        Either
+          .catchNonFatal(URI.create(uri))
```
chuwy, line 209: It in fact fails on very few sets invalid URIs, unfortunately, like `/` is valid. I think eventually we want to make this check more restrictive.

modules/beam/src/test/resources/enrichments_wrong/foo.json
```diff
@@ -0,0 +1 @@
+bar
```
chuwy, line 1: Don't think we need it?

chuwy (approved):
Looks great, just:

1. Leftover file - please delete iit
2. My comment in #388

modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/config.scala
benjben: Indeed we could. Should I at least check that it starts with `http://` or `https://` ?

modules/beam/src/test/resources/enrichments_wrong/foo.json
benjben: It's used here : https://github.com/snowplow/enrich/pull/392/files#diff-5e7b8cfab5c73bb6092007871ff982d471bdb1b42ac80a82350ab64669d87715R95
It's not directly related to this issue but I took the opportunity to improve unit testing to our configuration parsing

modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/config.scala
chuwy: I don't know. Probably. If we're sure Sentry supports HTTP(S) only.

modules/beam/src/test/resources/enrichments_wrong/foo.json
chuwy: Ah ok, sorry - I somehow thouoght it's in the repo's root.

modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/config.scala
benjben: Seems that this is a REST API : https://docs.sentry.io/api/requests/
I'm adding this additional check

oguzhanunlu (approved):
Thanks @benjben !

PR #403, opened 2020-12-08
Stream FS2: add a test specification with POST payload

chuwy:
Had to reproduce a test for RDB Shredder (https://github.com/snowplow/snowplow-rdb-loader/issues/238), thought it would be great to contribute back.

### DIFF
--- a/modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
+++ b/modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
@@ -13,7 +13,7 @@
 package com.snowplowanalytics.snowplow.enrich.fs2
 
 import java.time.Instant
-import java.util.UUID
+import java.util.{Base64, UUID}
 
 import scala.concurrent.duration._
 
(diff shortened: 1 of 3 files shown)


modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
```diff
@@ -80,6 +78,41 @@ class EnrichSpec extends Specification with CatsIO with ScalaCheck {
...
+          event_vendor = Some("com.snowplowanalytics.snowplow"),
+          event_name = Some("page_view"),
+          event_format = Some("jsonschema"),
+          event_version = Some("1-0-0"),
+          platform = Some("app"),
+          v_tracker = Some("test-0.0.1"),
+          contexts = Contexts(List(
+            SelfDescribingData(
+              SchemaKey("com.snowplowanalytics.snowplow", "change_form","jsonschema", SchemaVer.Full(1,0,0)),
+              json"""{"formId":"a","elementId":"b","nodeName":"TEXTAREA","value":"line 1\nline2\tcolumn2"}"""
```
benjben, line 96: Are you sure that putting `\n` in the string simulates a true new line in the JSON ?
```
scala> json"""{"foo":"b\nar"}""".noSpaces
res5: String = {"foo":"b\nar"}
```

benjben (approved):
:+1:

modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
chuwy: Yes! If it worked in your case making:

```json
{"foo":"b
ar"}
```

It would be an invalid JSON ([source](https://stackoverflow.com/questions/42068/how-do-i-handle-newlines-in-json)). `\\n` is what we do in input: https://github.com/snowplow/enrich/pull/403/files#diff-2ad55bdc3146219ba9dacfd9e3087e37c5a1947ce87eb0ad233eb0b67a5206dbR225

What is more important however is that when this enriched TSV gets shredded I can reproduce the exact error from https://github.com/snowplow/snowplow-rdb-loader/issues/238

modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/test/SchemaRegistry.scala
```diff
@@ -81,6 +81,108 @@ object SchemaRegistry {
 	    "additionalProperties": false
     }"""
 
+  // Defined on Iglu Central
```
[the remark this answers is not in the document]
chuwy: I did this to avoid any IO in the client, but still not sure if it was worth to do.

modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
```diff
@@ -80,6 +78,41 @@ class EnrichSpec extends Specification with CatsIO with ScalaCheck {
       }
     }
 
+    "enrich a minimal POST page_view CollectorPayload event with newline and without any enrichments enabled" in {
+      val expected = minimalEvent
+        .copy(
+          etl_tstamp = Some(Instant.ofEpochMilli(SpecHelpers.StaticTime)),
+          user_ipaddress = Some("175.16.199.0"),
```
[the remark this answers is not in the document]
chuwy: Yeah, you're right, maybe even extend the `minimalEvent` with it. Will do.

modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
benjben: I totally missed the fact that a JSON field can't be on several lines. Thanks

PR #417, opened 2021-03-23
FS2: StatsD reporter for metrics

istreeter:
This adds a custom dropwizard reporter -- which actually has more functionality that what is needed for Enrich! (e.g. histograms, timers).  But we could re-use the reporter for other apps, where we may need the extra functionality.  The new reporter is backed by the [dogstatsd client](https://github.com/DataDog/java-dogstatsd-client), which is non-blocking and runs on a dedicated thread.

There are some features I considered adding, but chose not to yet:
- Option to have both statsd and stdout reporters both running at the same time.
- Option to filter out some metrics from the reporter
- Option to configure the time unit of the metrics (seconds vs millis)

I also chose *not* to generate a random UUID to be a `worker_id` tag on the metrics.  I believe it is possible to add worker identifiers through configuration at run time.  For example, in a Kubernetes environment we can add the pod_name, which uniquely identifies the worker:
```
tags = {
  pod_name = ${HOSTNAME}
}
```

### DIFF
--- a/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/config/io.scala
+++ b/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/config/io.scala
@@ -18,6 +18,8 @@ import cats.syntax.either._
 
 import _root_.io.circe.{Decoder, Encoder}
 import _root_.io.circe.generic.extras.semiauto._
+import _root_.io.circe.config.syntax._
+import scala.concurrent.duration.FiniteDuration
 
 object io {
 
@@ -97,5 +99,27 @@ object io {
       }
     implicit val outputEncoder: Encoder[Output] =
       deriveConfiguredEncoder[Output]
+  }
+
+  sealed trait MetricsReporter {
+    def period: FiniteDuration
+  }
+
+  object MetricsReporter {
+    case class Stdout(period: FiniteDuration) extends MetricsReporter
+    case class StatsD(
+      period: FiniteDuration,
+      hostname: String,
+      port: Int,
+      tags: Map[String, String]
+    ) extends MetricsReporter
+
+    implicit val finiteDurationEncoder: Encoder[FiniteDuration] =
+      implicitly[Encoder[String]].contramap(_.toString)
+
+    implicit val metricsReporterDecoder: Decoder[MetricsReporter] =
+      deriveConfiguredDecoder[MetricsReporter]
+    implicit val metricsReporterEncoder: Encoder[MetricsReporter] =
+      deriveConfiguredEncoder[MetricsReporter]
   }
 }
--- a/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
+++ b/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
@@ -13,15 +13,17 @@
 package com.snowplowanalytics.snowplow.enrich.fs2.io
 
 import cats.syntax.applicativeError._
+import cats.Applicative
 import cats.effect.{Resource, Sync, Timer}
 
 import fs2.Stream
 
-import com.codahale.metrics.{Gauge, MetricRegistry, Slf4jReporter}
+import com.codahale.metrics.{Gauge, MetricRegistry, ScheduledReporter, Slf4jReporter}
 
 import org.slf4j.LoggerFactory
 
 import com.snowplowanalytics.snowplow.enrich.fs2.Environment
+import com.snowplowanalytics.snowplow.enrich.fs2.config.io.MetricsReporter
 
 trait Metrics[F[_]] {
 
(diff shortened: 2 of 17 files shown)


modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -47,10 +49,10 @@ trait Metrics[F[_]] {
...
 
   val LoggerName = "enrich.metrics"
-  val LatencyGaugeName = "enrich.metrics.latency"
-  val RawCounterName = "enrich.metrics.raw.count"
-  val GoodCounterName = "enrich.metrics.good.count"
-  val BadCounterName = "enrich.metrics.bad.count"
+  val LatencyGaugeName = "snowplow.enrich.metrics.latency"
+  val RawCounterName = "snowplow.enrich.metrics.raw"
+  val GoodCounterName = "snowplow.enrich.metrics.good"
+  val BadCounterName = "snowplow.enrich.metrics.bad"
```
istreeter, line 55: I am slightly worried it is a breaking change to rename these metrics.  On the other hand, fs2 is currently only released as "public beta", and until now metrics have been used only for debugging.  I think the new names are more appropriate when we start sending metrics to a statsd server.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala
```diff
@@ -0,0 +1,147 @@
+/*
+ * Copyright (c) 2020 Snowplow Analytics Ltd. All rights reserved.
```
chuwy, line 2: 2021

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/config/io.scala
```diff
@@ -98,4 +100,26 @@ object io {
...
+      hostname: String,
+      port: Int,
+      tags: Map[String, String]
+    ) extends MetricsReporter
+
+    implicit val finiteDurationEncoder: Encoder[FiniteDuration] =
+      implicitly[Encoder[String]].contramap(_.toString)
+
+    implicit val metricsReporterDecoder: Decoder[MetricsReporter] =
+      deriveConfiguredDecoder[MetricsReporter]
```
chuwy, line 121: I thought we were going to add a fallback option with `metricsReportPeriod` to make configuration foward-compatible?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/config/io.scala
```diff
@@ -98,4 +100,26 @@ object io {
     implicit val outputEncoder: Encoder[Output] =
       deriveConfiguredEncoder[Output]
   }
+
+  sealed trait MetricsReporter {
+    def period: FiniteDuration
+  }
+
+  object MetricsReporter {
+    case class Stdout(period: FiniteDuration) extends MetricsReporter
```
chuwy, line 109: Could be `final`

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -64,21 +66,24 @@ object Metrics {
...
+        for {
+          logger <- Resource.liftF(Sync[F].delay(LoggerFactory.getLogger(LoggerName)))
+          reporter <- Resource.fromAutoCloseable(Sync[F].delay(Slf4jReporter.forRegistry(registry).outputTo(logger).build))
+        } yield reporter
+      case statsd: MetricsReporter.StatsD =>
+        StatsDReporter.resource(statsd, registry)
     }
 
-  def make[F[_]: Sync](reporter: Slf4jReporter, registry: MetricRegistry): Metrics[F] =
+  def make[F[_]: Sync](reporter: ScheduledReporter, registry: MetricRegistry): Metrics[F] =
```
chuwy, line 86: But `ScheduledReporter` reports on its own in a background thread IIRC? So we can remove the `report` method?

chuwy (approved):
Looks great! Just not sure I understand the of scheduled reporter.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
istreeter: A `ScheduledReporter` would normally run on its own thread.  But we choose to run it on our own thread pool using the `report` method.

This is not a new change: the old `Sf4jReporter` is a subclass of `ScheduledReporter` - so we already had this setup.

What's different now.... the StatsD reporter uses a StatsD client underneath, and that client runs on its own thread.

I think this is a good overall setup - the client's blocking operations are run on a dedicated thread, but we do not waste an extra thread for our reporters which are non-blocking.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala
istreeter: Fixed in a follow-up commit

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -64,21 +66,24 @@ object Metrics {
...
-      val logger = LoggerFactory.getLogger(LoggerName)
-      val reporter = Slf4jReporter.forRegistry(registry).outputTo(logger).build()
-      (reporter, registry)
+  def resource[F[_]: Sync](config: MetricsReporter): Resource[F, Metrics[F]] =
+    for {
+      registry <- Resource.liftF(Sync[F].delay((new MetricRegistry())))
+      rep <- reporter(config, registry)
+    } yield make(rep, registry)
+
+  def reporter[F[_]: Sync](config: MetricsReporter, registry: MetricRegistry): Resource[F, ScheduledReporter] =
```
benjben, line 75: Maybe we could call `report` when closing the `Resource` ? Line 66 says so but I don't see it happening.

benjben (approved):

PR #418, opened 2021-03-23
Stream FS2: Add sink for Pii Events (close #416)

### LINKED ISSUE #416: Stream FS2: add sink for pii events
FS2 Enrich does not yet have a topic for publishing PII-containing events generated by the Pii Enrichment.

istreeter:
The primary purpose of this PR is:
* sink PII events into the dedicated PII topic.
* sink overly large messages into the bad topic

This PR includes refactoring some shared code into common, rather than have it implemented three times separately in beam, stream and fs2 modules.

I made a significant change to the `Payload` class - which needs some careful review.  Previously we were decomposing a list of events into a stream, replacing the "ack" with `F[Unit]` for all but the last event in the list.  I changed this to instead send the list through a pipe, and call the `finalise` only after all events are processed ([link](https://github.com/snowplow/enrich/pull/418/files#diff-7c8458a5ab1e0d1898535fbdcb94645e2824a1c1d76e7d928c71be6d2b2e5f40R30-R32))

```scala
  def sink[F[_]: Concurrent, A](inner: Pipe[F, A, Unit]): Pipe[F, Payload[F, A], Unit] =
    _.observeAsync(Enrich.ConcurrencyLevel)(_.map(_.data).through(inner))
      .evalMap(_.finalise)
```

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
@@ -27,6 +27,8 @@ import scala.util.control.NonFatal
 import cats.syntax.either._
 import cats.syntax.option._
 
+import com.snowplowanalytics.iglu.core.circe.implicits._
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer, SelfDescribingData}
 import com.snowplowanalytics.snowplow.badrows.{FailureDetails, Processor}
 import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.MiscEnrichments
@@ -34,7 +36,6 @@ import com.snowplowanalytics.snowplow.enrich.common.enrichments.MiscEnrichments
 import inet.ipaddr.HostName
 
 import io.circe.Json
-import io.circe.syntax._
 import io.lemonlabs.uri.{Uri, Url}
 import io.lemonlabs.uri.config.UriConfig
 import io.lemonlabs.uri.decoding.PercentDecoder
@@ -511,15 +512,26 @@ object ConversionUtils {
       }
 
   private def getContextParentEvent(eventId: String): Json =
-    Json.obj(
-      "schema" := Json.fromString("iglu:com.snowplowanalytics.snowplow/contexts/jsonschema/1-0-0"),
-      "data" := Json.arr(
-        Json.obj(
-          "schema" -> Json.fromString(
-            "iglu:com.snowplowanalytics.snowplow/parent_event/jsonschema/1-0-0"
-          ),
-          "data" -> Json.obj("parentEventId" -> Json.fromString(eventId))
-        )
-      )
-    )
+    SelfDescribingData(
+      SchemaKey("com.snowplowanalytics.snowplow", "contexts", "jsonschema", SchemaVer.Full(1, 0, 0)),
+      SelfDescribingData(
+        SchemaKey("iglu:com.snowplowanalytics.snowplow", "parent_event", "jsonschema", SchemaVer.Full(1, 0, 0)),
+        Json.obj("parentEventId" -> Json.fromString(eventId))
+      ).normalize
+    ).normalize
+
+  private val EnrichedFields =
+    classOf[EnrichedEvent].getDeclaredFields
+      .filterNot(_.getName.equals("pii"))
+      .map { field => field.setAccessible(true); field }
+      .toList
+
+  /** Format an EnrichedEvent as a TSV. */
+  def tabSeparatedEnrichedEvent(enrichedEvent: EnrichedEvent): String =
+    EnrichedFields
+      .map { field =>
+        Option(field.get(enrichedEvent)).getOrElse("")
+      }
+      .mkString("\t")
+
 }
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
@@ -27,10 +27,14 @@ import scala.util.control.NonFatal
 import cats.syntax.either._
 import cats.syntax.option._
 
-import com.snowplowanalytics.snowplow.badrows.FailureDetails
+import com.snowplowanalytics.snowplow.badrows.{FailureDetails, Processor}
+import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent
+import com.snowplowanalytics.snowplow.enrich.common.enrichments.MiscEnrichments
 
 import inet.ipaddr.HostName
 
+import io.circe.Json
+import io.circe.syntax._
 import io.lemonlabs.uri.{Uri, Url}
 import io.lemonlabs.uri.config.UriConfig
 import io.lemonlabs.uri.decoding.PercentDecoder
@@ -38,6 +42,8 @@ import io.lemonlabs.uri.encoding.percentEncode
 
 import org.apache.commons.codec.binary.Base64
 import org.apache.http.client.utils.URLEncodedUtils
+import org.joda.time.{DateTime, DateTimeZone}
+import org.joda.time.format.DateTimeFormat
 
 /** General-purpose utils to help the ETL process along. */
 object ConversionUtils {
@@ -477,4 +483,43 @@ object ConversionUtils {
   /** Extract valid IP (v4 or v6) address from a string */
   def extractInetAddress(arg: String): Option[InetAddress] =
     Option(new HostName(arg).asInetAddress)
+
+  private val eeTstampFormatter = DateTimeFormat.forPattern("yyyy-MM-dd HH:mm:ss.SSS")
+
+  /** Creates a PII event from the pii field of an existing event. */
+  def getPiiEvent(processor: Processor, event: EnrichedEvent): Option[EnrichedEvent] =
+    Option(event.pii)
+      .filter(_.nonEmpty)
+      .map { pii =>
+        val ee = new EnrichedEvent
+        ee.unstruct_event = pii
+        ee.app_id = event.app_id
+        ee.platform = "srv"
+        ee.etl_tstamp = event.etl_tstamp
+        ee.collector_tstamp = event.collector_tstamp
+        ee.event = "pii_transformation"
+        ee.event_id = UUID.randomUUID().toString
+        ee.derived_tstamp = eeTstampFormatter.print(DateTime.now(DateTimeZone.UTC))
+        ee.true_tstamp = ee.derived_tstamp
+        ee.event_vendor = "com.snowplowanalytics.snowplow"
+        ee.event_format = "jsonschema"
+        ee.event_name = "pii_transformation"
+        ee.event_version = "1-0-0"
+        ee.v_etl = MiscEnrichments.etlVersion(processor)
+        ee.contexts = getContextParentEvent(ee.event_id).noSpaces
+        ee
+      }
+
+  private def getContextParentEvent(eventId: String): Json =
+    Json.obj(
+      "schema" := Json.fromString("iglu:com.snowplowanalytics.snowplow/contexts/jsonschema/1-0-0"),
+      "data" := Json.arr(
+        Json.obj(
+          "schema" -> Json.fromString(
+            "iglu:com.snowplowanalytics.snowplow/parent_event/jsonschema/1-0-0"
+          ),
+          "data" -> Json.obj("parentEventId" -> Json.fromString(eventId))
+        )
+      )
+    )
 }
--- a/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
+++ b/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
@@ -60,7 +60,7 @@ object Enrich {
 
   /**
    * Run a primary enrichment stream, reading from [[Environment]] source, enriching
-   * via [[enrichWith]] and sinking into [[GoodSink]] and [[BadSink]] respectively.
+   * via [[enrichWith]] and sinking into the Good, Bad, and Pii sinks.
    * Can be stopped via _stop signal_ from [[Environment]]
    *
    * The stream won't download any enrichment DBs, it is responsibility of [[Assets]]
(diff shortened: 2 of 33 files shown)


modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Sinks.scala
```diff
@@ -86,24 +85,26 @@ object Sinks {
...
-        (s: Stream[F, Payload[F, A]]) => s.parEvalMapUnordered(Enrich.ConcurrencyLevel)(row => producer.produce(row.data) >> row.finalise)
-      )
+      .map(producer => (a: A) => producer.produce(a).void)
   }
 
-  def goodFileSink[F[_]: Sync: ContextShift](goodOut: Path, blocker: Blocker): GoodSink[F] =
-    goodStream =>
-      goodStream
-        .evalMap(p => p.finalise.as(Enrich.encodeEvent(p.data)))
+  def goodFileSink[F[_]: Concurrent: ContextShift](goodOut: Path, blocker: Blocker): GoodSink[F] =
```
chuwy, line 91: But it's one stream/file per event?

chuwy (commented):
Did a first pass. I guess the only real question is about file sink - to me they seem to br broken now as it should create a file per payload or I'm missing something?

chuwy:
> Responsibility is for acking is moved to Enrich, rather than the sink itself. This will reduce code duplication when we increase the number of Sinks we support, i.e. Kinesis.

My point about ack in stream was that it gives us most precise finalisation, i.e. we couldn't run into a situation when row has been processed, counted, acked, but then failed to be pushed into a stream. Whereas with sink-in-enrich we have a super small chance to lose an event in between ack and sink.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -69,16 +71,26 @@ object Enrich {
...
-      .void
+      .parEvalMapUnordered(ConcurrencyLevel)(sinkEither(badSink(env), goodSink(env)))
+  }
+
+  def sinkEither[F[_]: Applicative, A, B](sinkA: A => F[Unit], sinkB: B => F[Unit]): Payload.Parsed[F, A, B] => F[Unit] = {
+    case Left(parsed) => sinkA(parsed.data) *> parsed.finalise
+    case Right(parsed) => sinkB(parsed.data) *> parsed.finalise
+  }
+
+  def badSink[F[_]: Applicative](env: Environment[F])(br: BadRow): F[Unit] =
```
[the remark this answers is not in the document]
istreeter: 1. It's interesting to compare the two versions if you write `Pipe` out in full.  Which of these now looks more like a sink?
```
type Sink[F[_], A]  =  Stream[F, A]  =>  Stream[F, Unit]
type Sink[F[_], A]  =  A             =>  F[Unit]
```

I understand your arguments, but I find the latter *does* look like a sink.  At least as much as the Pipe version.

I did consider abstracting it a bit further with a cats `Arrow`, and then `type Sink[F[_, _], A] = F[A, Unit]`.  But ultimately I didn't see any advantage.

```
def badSink[F[_, _]: Arrow](sink: Sink[F])(br: BadRow): ??? // Haven't thought this through
```

2. You're right - I should be consistent.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Sinks.scala
istreeter: Will fix this.

istreeter (commented):
> My point about ack-in-sink was that it gives us most precise finalisation, i.e. we couldn't run into a situation when row has been processed, counted, acked, but then failed to be pushed into a stream. Whereas with sink-in-enrich we have a super small chance to lose an event in between ack and sink.

The problem I had was around acking events with PII.  The order of actions with no PII is:

1. Enrich event
2. Sink event to good
3. Ack event

The order of actions with PII is:

1. Enrich event
2. Sink event to good *but do not ack it yet*
3. Extract PII event
4. Sink PII event
5. Ack event

I could not find a tidy way to implement both flows, so long as the ack happened in the stream.  But with ack-in-enrich it is possible and quite neat.

> Whereas with sink-in-enrich we have a super small chance to lose an event in between ack and sink.

I don't see how this can happen.  The sink `F[Unit]` has to run all the way to success before the ack is attempted.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -69,16 +71,21 @@ object Enrich {
...
-    val goodSink: GoodSink[F] = _.evalTap(_ => env.metrics.goodCount).through(env.good)
 
     env.source
       .pauseWhen(env.pauseEnrich)
       .evalTap(_ => env.metrics.rawCount)
       .parEvalMapUnordered(ConcurrencyLevel)(enrich)
       .flatMap(_.decompose[BadRow, EnrichedEvent])
-      .observeEither(badSink, goodSink)
-      .void
+      .parEvalMapUnordered(ConcurrencyLevel)(Payload.sinkEither(badSink(env), goodSink(env)))
```
istreeter, line 80: TODO: I think this should not be parallelised.  Payloads should sunk in order, so the final ack means all payloads of a batch have been sunk.

benjben: > so the final ack means all payloads of a batch have been sunk.

This would be true with Kinesis where we ack on offset, and not a message. I wonder if this is also true for PubSub.

We already have `parEvalMapUnordered` before, so I think that we already lost the order of messages at this point.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
```diff
@@ -122,6 +124,7 @@ object Environment {
         metrics <- Metrics.resource[F]
         rawSource = Source.read[F](blocker, file.auth, file.input)
         goodSink <- Sinks.goodSink[F](blocker, file.auth, file.good)
+        piiSink <- file.pii.map(f => Sinks.goodSink[F](blocker, file.auth, f)).getOrElse(Resource.pure[F, GoodSink[F]](_ => Sync[F].unit))
```
benjben, line 127: Could be nice to put what's inside the `getOrElse` into a `noop` sink.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
istreeter: My comment refers to PubSub too, where we ack every message.  This is specifically for messages that comprise multiple events.

Line 79 decomposes a each message into stream of multiple events:
```scala
.flatMap(_.decompose[BadRow, EnrichedEvent])
```
The `decompose` method sets things up so that the final event in the batch (for a single pubsub message) will ack the message.  But this only works if we process the batch of events in order.

benjben: I get it, thanks !

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Channel.scala
```diff
@@ -0,0 +1,100 @@
...
+
+  /** Sends a bad/good/pii item into the appropriate sink */
+  def sink[F[_]: Concurrent, B, G, P](
+    bad: Pipe[F, B, Unit],
+    good: Pipe[F, G, Unit],
+    pii: Pipe[F, P, Unit]
+  ): Pipe[F, Channel[B, G, P], Unit] =
+    _.observe(_.collect { case Bad(br) => br }.through(bad))
+      .observe(_.collect { case Good(event) => event }.through(good))
+      .observe(_.collect { case Pii(event) => event }.through(pii))
```
benjben, line 50: I'm not at ease with the fact that the type of each element inside the `Stream` of `Channel` can change, and that we need to call 3 times `observe()` and use pattern matching on class types.

I would personally be in favor of having more granular control over the `Stream`s and not have a generic `Channel` that can wrap any type of data.

That might just be a matter of taste, I'm not sure about performance benefits.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Channel.scala
```diff
@@ -0,0 +1,100 @@
...
+
+import cats.effect.Concurrent
+import fs2.Pipe
+
+import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent
+import com.snowplowanalytics.snowplow.badrows.{BadRow, Failure, Payload => BadRowPayload}
+import com.snowplowanalytics.snowplow.enrich.common.utils.ConversionUtils
+
+/** An item that can be sunk into any one of the enrichment sinks: Bad, Good, or Pii */
+sealed trait Channel[+Bad, +Good, +Pii]
```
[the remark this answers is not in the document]
chuwy: Agree with Ben, especially given FS2 3 has new `Channel` type: https://github.com/typelevel/fs2/pull/2326. I think something like `Output` would fit better.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Channel.scala
istreeter: Let me try to make you feel more at ease with this!

Channel (I will rename it to Output) is basically a 3-way `Either`.  While `Either` has just `Left` and `Right`, our `Output` has 3 possible types `Bad`, `Good`, `Pii` - but apart from that they are the same thing.

Now FS2 has a function called `observeEither` [which is implemented exactly the same as above](https://github.com/typelevel/fs2/blob/2af0571de34cc11b6b2bafeef0baf487f0760ad8/core/shared/src/main/scala/fs2/Stream.scala#L3672-L3681).  It does pattern matching on the elements, and it calls `observe` for each possible pattern match.  I think this is the only way to direct a stream into several different sinks depending on the result of processing.

> I would personally be in favor of having more granular control over the Streams

I felt the same!  This was why I experimented with typing the sinks as `A => F[Unit]` instead of typing them as `Pipe[F, A, Unit]`.  I found the former let me have more control over exactly when to sink events; whereas the latter forced me to use this multi-`observe` pattern and the 3-way either.

Unfortunately my experiment with `A => F[Unit]` failed because things got messy with the file sink.

Can you think of any other way of send a stream of 3 different outputs into 3 different sinks?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Channel.scala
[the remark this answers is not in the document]
istreeter: I really struggled to choose a good name for this trait.

What do you think about calling it `Either3` as that's what is really is?  I could call the sub-classes `Left`, `Middle` and `Right`.

Or do you prefer `Output` with sub-classes `Bad`, `Good`, and `Pii`?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Channel.scala
benjben: Thank you for the link for `observeEither` implementation, your implementation should be the optimal one then.

> Can you think of any other way of send a stream of 3 different outputs into 3 different sinks?

I can't. I guess the alternative would be to not have this `Output` class and to use multiple `observeEither` after each step (enrich, serialization with possibly size violation), but your approach is probably easier to read.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -12,30 +12,89 @@
...
+/** An item that can be sunk into any one of the enrichment sinks: Bad, Good, or Pii */
+sealed trait Output[+Bad, +Good, +Pii]
 
 object Output {
 
-  case class Bad(badRow: BadRow) extends Output
-  case class Good(event: EnrichedEvent) extends Output
-  case class Pii(event: EnrichedEvent) extends Output
+  /** An item destined for the Bad Sink */
+  case class Bad[B](badRow: B) extends Output[B, Nothing, Nothing]
```
chuwy, line 31: Can be `final`

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -12,30 +12,89 @@
...
+    case g: Good[G] => g
+    case p: Pii[P] => p
+  }
+
+  private def serializeEnriched(enriched: EnrichedEvent): Either[BadRow, Array[Byte]] = {
+    val asStr = ConversionUtils.tabSeparatedEnrichedEvent(enriched)
+    val asBytes = asStr.getBytes(UTF_8)
+    val size = asBytes.length
+    if (size > MaxRecordSize) {
+      val msg = "event passed enrichment but exceeded the maximum allowed size as a result"
```
chuwy, line 85: Worth mentioning `MaxRecordSize`?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -12,30 +12,89 @@
...
+        .SizeViolation(
+          Enrich.processor,
+          Failure.SizeViolation(Instant.now(), MaxRecordSize, size, msg),
+          BadRowPayload.RawPayload(asStr.take(MaxMessageSize))
+        )
+      Left(br)
+    } else Right(asBytes)
+  }
+
+  private val MaxRecordSize = 6900000
```
chuwy, line 96: Also probably worth mentioning in a comment why it's `6900000` while official is something around 10Mb

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -12,30 +12,89 @@
...
+   * Serializes events and bad rows into byte arrays. Good/Pii events exceeding
+   *  a maximum size after serialization are converted to bad events.
+   */
+  val serialize: Parsed => Serialized =
+    serializeGood.andThen(serializePii).andThen(serializeBad)
+
+  private def serializeGood[P]: Output[BadRow, EnrichedEvent, P] => Output[BadRow, Array[Byte], P] = {
+    case Good(g) =>
+      serializeEnriched(g).fold(Bad(_), Good(_))
+    case b: Bad[BadRow] => b
```
chuwy, line 63: I think mentioning a generic type in pattern match might be a bit confusing for people not knowing about type erasure.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ConversionUtils.scala
```diff
@@ -477,4 +483,43 @@ object ConversionUtils {
...
+        ee.event_name = "pii_transformation"
+        ee.event_version = "1-0-0"
+        ee.v_etl = MiscEnrichments.etlVersion(processor)
+        ee.contexts = getContextParentEvent(ee.event_id).noSpaces
+        ee
+      }
+
+  private def getContextParentEvent(eventId: String): Json =
+    Json.obj(
+      "schema" := Json.fromString("iglu:com.snowplowanalytics.snowplow/contexts/jsonschema/1-0-0"),
```
chuwy, line 515: I'd probably use `SelfDescribingData` with serialization here. Just in sake of having `SchemaKey("com.snowplowanalytics.snowplow", "parent_event", "jsonschema", SchemaVer.Full(1,0,0)` constant defined somewhere. But it's very stylistic.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright (c) 2020 Snowplow Analytics Ltd. All rights reserved.
```
chuwy, line 2: 2021

chuwy (approved):
Looks good - just few stylistic points (feel free to ignore), but let me do another pass - started reviewing commit-by-commit and I think something might get lost.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
istreeter: Will change to

```scala
case b @ Bad(_) => b
case p @ Pii(_) => p
```

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
istreeter: I will mention it is because of base64 encoding.

By the way, I would have guessed the max cut off would be closer to 7500000.  Because base64 encoding has a 4/3 expansion factor.

PR #426, opened 2021-04-08
Stream FS2: Add message attributes to PubSub (close #340)

### LINKED ISSUE #340: Stream FS2: add message attributes to PubSub
**Project**: Beam Enrich

**Version**: N/A

**Expected behavior**: Message are published with some attributes (e.g., app_id, event_name)

**Actual behavior**: N/A

It's possible to publish messages with high level attributes such as app_id and event_name to the enriched events themselves. This would add the ability for downstream subscribers to filter events making it easier to selectively subscribe to events or for multitenancy applications (e.g., sinking app_ids to different BQ datasets / tables).

istreeter:


config/config.fs2.hocon.sample
```diff
@@ -18,6 +18,9 @@ good = {
   type = "PubSub"
   topic = "projects/test-project/topics/good-topic"
 
+  // Enriched event fields to add as PubSub message attributes.
```
chuwy, line 21:
```suggestion
  // Enriched event fields to add as PubSub message attributes (optional)
```

Also, a duplicate of below?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/AttributedData.scala
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright (c) 2020 Snowplow Analytics Ltd. All rights reserved.
```
chuwy, line 2: 2021

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
```diff
@@ -90,21 +90,41 @@ object Enrich {
...
+  def addAttributes(fieldNames: Set[String]): EnrichedEvent => AttributedData[EnrichedEvent] = {
+    val fields = ConversionUtils.EnrichedFields.filter(f => fieldNames.contains(f.getName)).map(f => f.getName -> f).toMap
+
+    ee => {
+      val attrs = fields
+        .map {
+          case (k, f) => k -> Option(f.get(ee))
+        }
+        .collect {
+          case (k, Some(v)) => k -> v.toString
```
istreeter, line 123: TODO: Filter strings that exceed 1024 bytes

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/AttributedData.scala
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the Apache License Version 2.0 is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.snowplow.enrich.fs2
+
+/** Data intended to be sent to a sink, plus key-value pairs to be added to the message as attributes */
+final case class AttributedData[A](data: A, attributes: Map[String, String])
```
chuwy, line 16: A very high-level note: `AttributedData` feels very related to `Payload`. I.e. `Payload` can be not just data + checkpoint, but data + checkpoint + metadata. Obviously not proposing to combine them now, but just saying this in the light of "FS2 data processing framework" idea - this primitive would be super common.

chuwy (commented):
A very quick first pass.

config/config.fs2.hocon.sample
benjben: Below is for pii events

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala
chuwy: I think we should trim, not filter out. Also just in case, there are very few fields that can be that long: https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L123-L157

PR #428, opened 2021-04-13
Stream FS2: StatsD reporter for metrics (close #421)

### LINKED ISSUE #421: Stream FS2: statsd reporter for metrics
FS2 Enrich currently uses Dropwizard metrics and reports them to stdout.

The stdout solution is suitable for development only; a better setup for production is to send metrics to a StatsD server.  The StatsD server is then responsible for forwarding metrics to any backend.

istreeter:
This the re-worked version.  We no longer use a 3rd party statsd client, but rather implement the statsd protocol ourselves.  This is because I could not find a 3rd party client that respected DNS changes during a long-running process.

Here, I call `InetAddress.getByName()` every time I want to send metrics.  This gives the JVM opportunity to re-query the DNS server.  The JVM will respect the ttl on DNS records, but otherwise will cache the result of a DNS lookup.

### DIFF
--- a/config/config.fs2.hocon.sample
+++ b/config/config.fs2.hocon.sample
@@ -52,6 +52,30 @@ sentry = {
 // no assets will be updated if the key is absent
 assetsUpdatePeriod = "7 days"
 
-// Optional, period after Dropwizard will print out its metrics
-// no metrics will be printed if the key is absent
-metricsReportPeriod = "10 seconds"
+// Optional, configure how metrics are reported
+metrics = {
+
+  // Send metrics to a StatsD server on localhost
+  type = "StatsD"
+  hostname = "localhost"
+  port = 8125
+
+  // Alternatively, log to stdout using Slf4j
+  // type = "Stdout"
+
+  // Required, how frequently to report metrics
+  period = "10 seconds"
+
+  // Any key-value pairs to be tagged on every metric
+  tags = {
+    app = enrich
+
+    // Environment variables can be resolved to tags
+    // host = ${HOSTNAME}
+  }
+
+  // Optional, override the default metric prefix
+  // prefix = "snowplow.enrich."
+
+
+}
--- a/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
+++ b/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
@@ -59,7 +59,6 @@ import com.snowplowanalytics.snowplow.enrich.fs2.io.{FileSystem, Metrics, Sinks,
  * @param sentry              optional sentry client
  * @param metrics             common counters
  * @param assetsUpdatePeriod  time after which enrich assets should be refresh
- * @param metricsReportPeriod period after which metrics are updated
  */
 final case class Environment[F[_]](
   igluClient: Client[F, Json],
@@ -73,8 +72,7 @@ final case class Environment[F[_]](
   bad: ByteSink[F],
   sentry: Option[SentryClient],
   metrics: Metrics[F],
-  assetsUpdatePeriod: Option[FiniteDuration],
-  metricsReportPeriod: Option[FiniteDuration]
+  assetsUpdatePeriod: Option[FiniteDuration]
 )
 
 object Environment {
@@ -121,7 +119,7 @@ object Environment {
       for {
         client <- Client.parseDefault[F](parsedConfigs.igluJson).resource
         blocker <- Blocker[F]
-        metrics <- Metrics.resource[F]
+        metrics <- Resource.liftF(metricsReporter[F](blocker, file))
         rawSource = Source.read[F](blocker, file.auth, file.input)
         goodSink <- Sinks.sink[F](blocker, file.auth, file.good)
         piiSink <- file.pii.map(f => Sinks.sink[F](blocker, file.auth, f)).sequence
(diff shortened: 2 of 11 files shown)


modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala
```diff
@@ -0,0 +1,165 @@
...
+ * We use the DogStatsD extension to the StatsD protocol, which adds arbitrary key-value tags to the metric, e.g:
+ * `snowplow.enrich.good.count:20000|g|#app_id:12345,env:prod`
+ */
+object StatsDReporter {
+
+  /**
+   * A stream which periodically sends metrics from the registry to the StatsD server.
+   *
+   * The stream refreshes the InetAddress of the server each time it sends metrics, instead of
+   * caching the inet address, or keeping a long-running channel open. This allows us to
```
benjben, line 43: Can you confirm that even though we are calling `InetAddress.getByName()` each time we want to send metrics, there is some caching and the DNS request is performed only when the TTL has expired ?It would be great to add it in the comment.

config/config.fs2.hocon.sample
```diff
@@ -52,6 +52,31 @@ sentry = {
...
+
+    // Environment variables are resolved to tags
+    // host = ${HOSTNAME}
+  }
+
+
+  // The default reporter if metricsReporter is not configured
+  // Suitable for development
+  // type = "Stdout"
+}
```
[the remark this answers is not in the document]
istreeter: In a previous version of this PR I had it set up exactly as you describe!

But then we talked about making the configuration backwards compatible with version 1.4.2 (see [here](https://github.com/snowplow/enrich/blob/81d108152b54961867cd1f6218b22465afb5c083/config/config.fs2.hocon.sample#L45-L47)), and for that reason I moved `metricsReportPeriod` back outside of the `metrics` block.  But I agree it feels misplaced now.

What do you think?  Is it acceptable to break backwards compatibility in order to make the config sane, and the code simpler?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala
istreeter: The details are in [the InetAddress docs](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/net/InetAddress.html) in the section starting "InetAddress Caching".  The important line is

> The default setting is to cache for an implementation specific period of time.

Meaning it depends on our JVM, and on security settings.  I have tested this locally and I found it resolved a new IP address very quickly.  But in the version we deploy to kubernetes we might find it behaves differently - I have no idea.  We need to test it.

I can put some of these details in the comment, as you suggest.

istreeter: I found the answer in the `java.security` file of our docker image:

```
# The Java-level namelookup cache policy for successful lookups:
#
# any negative value: caching forever
# any positive value: the number of seconds to cache an address for
# zero: do not cache
#
# default value is forever (FOREVER). For security reasons, this
# caching is made forever when a security manager is set. When a security
# manager is not set, the default behavior in this implementation
# is to cache for 30 seconds.
#
# NOTE: setting this to anything other than the default value can have
#       serious security implications. Do not set it unless
#       you are sure you are not exposed to DNS spoofing attack.
#
#networkaddress.cache.ttl=-1
```

So for us it is 30 seconds.

config/config.fs2.hocon.sample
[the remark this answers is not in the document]
benjben: Given that we have full control over the Terraform used to deploy I don't see an issue with breaking backward compatibility in the configuration. For OSS users we can make it clear in the release notes.

In the discussion that you had did it come as a requirement to maintain backward compatibility ?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala
benjben: Awesome, thank you for the details!

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
```diff
@@ -159,6 +156,9 @@ object Environment {
   def makePause[F[_]: Concurrent]: Resource[F, SignallingRef[F, Boolean]] =
     Resource.make(SignallingRef(true))(_.set(true))
 
+  private def metricsReporter[F[_]: Sync: ContextShift: Timer](blocker: Blocker, config: ConfigFile): F[Metrics[F]] =
+    config.metrics.map(Metrics.build[F](blocker, _)).getOrElse(Metrics.noop[F].pure[F])
+
```
benjben, line 161:
```suggestion
  private def metricsReporter[F[_]: Sync: ContextShift: Timer](blocker: Blocker, config: ConfigFile): F[Metrics[F]] =    
    Metrics.build[F](blocker, config.metrics)
```

Just a proposal. That's how I do it for [RDB loader](https://github.com/snowplow/snowplow-rdb-loader/blob/feature/metrics/modules/loader/src/main/scala/com/snowplowanalytics/snowplow/rdbloader/dsl/StatsD.scala#L40-L58). I don't know what the best approach is, it might be just a matter of taste.

benjben (approved):
:+1: 

Really cool to have updated the config to look like this!

PR #431, opened 2021-04-14
Common: Bump commons-codec to 1.15 (close #430)

### LINKED ISSUE #430: Common: bump commons-codec to 1.15
A unit test in conversionUtilsSpec failed on a randomly generated string generated by scalacheck:

```
Error:    x decodeBase64Url should not return failure on any other string
Error:   Falsified after 70 passed tests.
Error:   > ARG_0: "3u"
Error:   > ARG_0_ORIGINAL: "鸯鋗隖ᣰኢ㮈◳3ꇞ著檐亅윍㏐锰鐚묻䷰㒾哫蠵盜嚛඼嚪䷤隊ኼਸ흂㗙㬍漰⸷ﴎ쨗櫤憠껖낧uႇ↳㠤鸽、ꑶ༓Ʉ髴㨯漩躶"
Error:   The seed is pwY8zvcMyQEEqruPZQJpw0ju_x9ZUYzRi9xnRifCu3L=
Error:   
Error:   > Left(could not base64 decode: Last encoded character (before the paddings if any) is a valid base 64 alphabet but not a possible value) is not Right (conversionUtilsSpecs.scala:220)
```

It is probably related to [this explanation](https://issues.apache.org/jira/browse/CODEC-263?focusedCommentId=16981194&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-16981194) in Apache's jira.  Behaviour changed in version 1.13 of the codec.

istreeter:


chuwy (approved):
As I mentioned in https://github.com/snowplow/enrich/issues/430#issuecomment-819830386, `commons-codec` seems to be not preserving compatibility between versions (at least spotted a couple of time before) and thus I'd be cautious, but if there's no way - I'm fine with it. Also it's been likely bumped transitively in some of the assets.

PR #433, opened 2021-04-15
Stream FS2: Run blocking enrichments on a separate execution context …

### LINKED ISSUE #432: Stream FS2: run blocking enrichments on a separate execution context
Currently, some enrichments block the threads (e.g. ApiRequestEnrichment).  This is especially bad for FS2 enrich, where the main thread pool is kept deliberately small.  To fix this problem, we need to pass around a `cats.effect.Blocker` so we block on a separate thread pool.  We need some indirection because Beam/Kinesis (etc) apps use the `Id` effect type, for which a ContextShift is not available.

istreeter:
…(close #432)

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -216,6 +216,7 @@ lazy val fs2 = project
       Dependencies.Libraries.log4cats,
       Dependencies.Libraries.catsRetry,
       Dependencies.Libraries.http4sClient,
+      Dependencies.Libraries.http4sCirce,
       Dependencies.Libraries.fs2BlobS3,
       Dependencies.Libraries.fs2BlobGcs,
       Dependencies.Libraries.metrics,
--- a/modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/singleton.scala
+++ b/modules/beam/src/main/scala/com.snowplowanalytics.snowplow.enrich.beam/singleton.scala
@@ -19,6 +19,7 @@ import cats.syntax.either._
 import com.snowplowanalytics.iglu.client.Client
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.EnrichmentRegistry
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.registry.EnrichmentConf
+import com.snowplowanalytics.snowplow.enrich.common.utils.BlockerF
 import io.circe.Json
 
 /** Singletons needed for unserializable classes. */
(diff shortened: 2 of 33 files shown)


benjben (approved):
It looks great @istreeter !

I wonder if we should also use the `Blocker` to valide SDJs (e.g. [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala#L56) and [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala#L169)) because it requires to communicate with Iglu server. But there is some caching in Iglu client so the communication does not need to happen for each event. @chuwy is it possible to provide Iglu client with the `Blocker` and it uses it only when it need to?

chuwy (approved):
Looks great. I was thinking more about a capability (implicit value), but also agree it's not cool to pass around mutable resources (like `Blocker`) as implicit values.

chuwy:
> @chuwy is it possible to provide Iglu client with the Blocker and it uses it only when it need to?

Sorry, missed it @benjben.

As I mentioned this morning - it's possible to do this in an efficient way only if we extrat this capability as a library and include into Iglu Client. Otherwise we can do this in an inefficient way by wrapping both impure (HTTP) and pure (cache) lookups and that could actually hurt performance in cases where most of the hits are cache-backed, because all those lookups will end up as unnecessary context switches.

If we extract this as a micro-library we could use it in other client libs that potentially do cache-backed IO lookups and have to be used in Beam/Spark.

benjben:
> that could actually hurt performance in cases where most of the hits are cache-backed, because all those lookups will end up as unnecessary context switches

Yes, what I wanted to know is if it's possible to give kind of an `Option[Blocker]` to Iglu client when we initialize it and then if present it uses it when doing the HTTP requests. If I understand correctly that's not supported yet.

istreeter:
I think this is definitely the way we want to head.  Especially whilst we are supporting some apps using the `Id` effect, and other `IO` apps whose preference is to run on a minimum thread pool.

The only question is..... do we go ahead and make this change to iglu client before we release Enrich 2.0.0?  The code change probably won't take long, but there's a lot of apps that use iglu client as dependency.

> Yes, what I wanted to know is if it's possible to give kind of an Option[Blocker] to Iglu clien

Just a small detail, I would probably make it take a `Blocker` not a `Option[Blocker]` but then provide a `Blocker.noop` as the equivalent of `None`.

PR #436, opened 2021-04-16
Stream FS2 2.0.0

istreeter:
This PR comprises three changes for FS2 Enrich

- Add message attributes to pubsub
- Tune source and sink configs to optimise throughput
- Reset latency metric to zero when there are no more events.


modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -53,14 +56,25 @@ object Metrics {
   val GoodCounterName = "good"
   val BadCounterName = "bad"
 
-  def build[F[_]: ContextShift: Sync: Timer](
+  /**
+   * Controls how quickly the latency gauge is reset to zero if there are no new incoming events.
+   * The gauge is reset after 5 * the period of reporting.
```
benjben, line 61: Why 5* and not 1* ?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -119,11 +126,25 @@ object Metrics {
 
       def badCount: F[Unit] =
         Sync[F].delay(badCounter.inc())
+    }
 
-      private def getGauge(now: Long, collectorTstamp: Long): Gauge[Long] =
-        new Gauge[Long] {
-          def getValue: Long = now - collectorTstamp
+  def latencyGauge[F[_]: Clock: Effect](config: MetricsReporter, ref: Ref[F, Option[(Long, Long)]]): Gauge[Long] = {
+    val resetDelta = config.period.toMillis * GaugeResetFactor
```
benjben, line 132: What about resetting the latency when we report the metrics ?

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Metrics.scala
```diff
@@ -119,11 +126,25 @@ object Metrics {
...
+    effectGauge(ref.get) {
+      case Some((acked, collectorTstamp)) =>
+        Clock[F].realTime(MILLISECONDS).map { now =>
+          if (now - acked < resetDelta)
+            acked - collectorTstamp
+          else
+            0L
```
benjben, line 139: Do we really want to send a latency of 0 ? I would be in favor of not sending it instead (maybe by filtering out [here](https://github.com/snowplow/enrich/blob/streeter/dev/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/StatsDReporter.scala#L87-L92)).

istreeter (commented):
Thanks for the review @benjben.  I changed things up a bit, so now the latency is reported only if it is defined, but it is reset after it has been reported.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
```diff
@@ -121,33 +124,35 @@ object Environment {
       for {
         client <- Client.parseDefault[F](parsedConfigs.igluJson).resource
         blocker <- Blocker[F]
-        metrics <- Metrics.resource[F]
+        metrics <- Resource.liftF(metricsReporter[F](blocker, file))
```
chuwy, line 127: It seems we're still not using latest CE2. `liftF` has been deprecated there.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
```diff
@@ -68,13 +70,14 @@ final case class Environment[F[_]](
...
-  good: ByteSink[F],
-  pii: Option[ByteSink[F]],
+  good: AttributedByteSink[F],
+  pii: Option[AttributedByteSink[F]],
   bad: ByteSink[F],
   sentry: Option[SentryClient],
   metrics: Metrics[F],
   assetsUpdatePeriod: Option[FiniteDuration],
-  metricsReportPeriod: Option[FiniteDuration]
+  goodAttributes: Set[String],
```
chuwy, line 79: Do we have any validation logic, checking that `foobar` is not possible? I was thinking about switching to `Set[Field]` at first, but also feel uncomfortable about polluting Scala code with something reflection-related.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Output.scala
```diff
@@ -36,18 +36,20 @@ object Output {
...
   /** Sends a bad/good/pii item into the appropriate sink */
   def sink[F[_]: Concurrent, B, G, P](
     bad: Pipe[F, B, Unit],
     good: Pipe[F, G, Unit],
     pii: Pipe[F, P, Unit]
   ): Pipe[F, Output[B, G, P], Unit] =
-    _.observe(_.collect { case Bad(br) => br }.through(bad))
-      .observe(_.collect { case Good(event) => event }.through(good))
-      .observe(_.collect { case Pii(event) => event }.through(pii))
+    _.observeAsync(Int.MaxValue)(_.collect { case Good(event) => event }.through(good))
```
chuwy, line 50: Are you sure we need `observeAsync` here? I'm starting to suspect this might be a performance problem we're seeing.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/io/Sinks.scala
```diff
@@ -30,19 +30,41 @@ import com.permutive.pubsub.producer.Model.{ProjectId, Topic}
...
+   * DelayThreshold should not ever cause blocking.
    */
   val DelayThreshold: FiniteDuration = 200.milliseconds
 
+  /**
+   * A batch of messages will be emitted to PubSub when the batch reaches 1000 messages.
+   * We use 1000 because it is the maximum batch size allowed by PubSub.
+   * This overrides the permutive library default of `5`
+   */
+  val PubsubMaxBatchSize = 1000L
```
chuwy, line 53: Eventually we'll have to make all these properties configurable. Beam allows us to configure them and it saved a day couple of times already.

chuwy (approved):
Looks great! Just a couple of questions.

modules/fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/fs2/EnrichSpec.scala
```diff
@@ -174,6 +175,99 @@ class EnrichSpec extends Specification with CatsIO with ScalaCheck {
...
+        for {
+          _ <- input.through(Enrich.resultSink(environment)).compile.drain
+          _ <- test.finalise()
+          good <- test.good.dequeue.compile.toList
+          pii <- test.pii.dequeue.compile.toList
+          bad <- test.bad.dequeue.compile.toList
+        } yield {
+
+          good should beLike {
+            case AttributedData(bytes, attrs) :: Nil =>
```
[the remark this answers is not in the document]
istreeter: This is the error if I change the test so that `good.size > 1`:

```
[error]   x emit a pii event with attributes to the pii sink
[error]    List(AttributedData([B@70e84169,Map(app_id -> test_app)), AttributedData([B@45696b98,Map(app_id -> test_app))) is incorrect:  (EnrichSpec.scala:227)
```

There is no `MatchError` because `beLike` takes a partial function, not a total function.

modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Environment.scala
istreeter: I do not validate names, but I ignore invalid names.  [This is the relevant line](https://github.com/snowplow/enrich/blob/9bdd973092716e61d57fe77d034f03c95e0e41f6/modules/fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/fs2/Enrich.scala#L115)

chuwy: I have a slight preference towards failing on unexpected names - I think silent skip can lead to a situation where user has a typo and expects certain behavior, but not getting it. Also fail-fast is more consistent with enrichments parsing logic.

PR #464, opened 2021-04-30
Stream FS2: Rename fs2-enrich to snowplow-enrich-pubsub

### LINKED ISSUE #463: Stream FS2: rename fs2-enrich to snowplow-enrich-pubsub
We make this naming decision because `snowplow-enrich` will be the future single asset that should replace all other versions of enrich.  The fact that is uses fs2 is an implementation detail rather than a feature.

istreeter:


oguzhanunlu (approved):

chuwy (commented):
I'm not pushing back on this at all, just want to make sure we took all pros and cons into account and understand how it's going to evolve after "Enrich FS2" becomes dominant.

* How confident are we that nobody will need Beam (or Spark or Flink) Enrich again?
* If it's going to be a single `snowplow-enrich` - how would we handle different sources and destinations: PubSub, Kinesis, RabbitMQ, Kafka etc. Are these going to be different components or a single one? If different, what naming convention would look like?

oguzhanunlu:
@chuwy pr is ready for review, publishing 2.0.0-rc1 [failed](https://github.com/snowplow/enrich/runs/2513466693?check_suite_focus=true#step:6:200) only for `snowplow-enrich-pubsub` with `denied: requested access to the resource is denied`. I think we need to create a new repository in docker hub for the new name. Could you do that?

chuwy (changes requested):
Hey @oguzhanunlu! One last file: `config/config.fs2.hocon.sample`

chuwy (changes requested):
Hey @oguzhanunlu. One last file: `config/config.fs2.hocon.sample`

chuwy (approved):
Looks great!

PR #474, opened 2021-07-05
Stream FS2: split into common module and PubSub module (close #473)

### LINKED ISSUE #473: enrich-pubsub: split into common-fs2 module and PubSub module
In order to add Kinesis support, some refactoring is needed to put the common code for PubSub Kinesis and Kafka into a common module and then have one module for each stack with the dependency.

For now we want to keep Stream Enrich in the repo.

benjben:
Enrich PubSub split into common_fs2 module and common module.

Before, `Environment.scala` was in charge of
- parsing the config files
- initializing PubSub source and sinks
- initializing other resources (e.g. Iglu client)

Now the logic is as follows:
- PubSub `main()` parses the configuration with `ParsedConfigs.parse[IO](cfg)`. The configuration format will be common to all assets and is therefore in `common_fs2`
- then the `main()` initializes PubSub source and sinks based on this configuration
- this source and the sinks are then passed to `Environment`

Case class `Environment` is the same as before.

Even if a sink doesn't support attributes (e.g. Kinesis), `AttributedByteSink` will be used for all assets.


build.sbt
```diff
@@ -200,26 +200,23 @@ lazy val beam =
 
 Global / onChangedBuildSource := ReloadOnSourceChanges
 
-lazy val pubsub = project
-  .in(file("modules/pubsub"))
+lazy val commonFs2 = project
+  .in(file("modules/common_fs2"))
```
chuwy, line 204: It's slightly unusual to use underscores in module names, hyphen should be used.

modules/common_fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -55,7 +61,7 @@ object Enrich {
   /** Default adapter registry, can be constructed dynamically in future */
   val adapterRegistry = new AdapterRegistry()
 
-  val processor: Processor = Processor(generated.BuildInfo.name, generated.BuildInfo.version)
+  val processor: Processor = Processor(BuildInfo.name, BuildInfo.version)
```
chuwy, line 64: I think we'd get something like `Processor("common-fs2", "2.0.0")`, which is not exactly what we want.

modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Main.scala
```diff
@@ -47,14 +52,22 @@ object Main extends IOApp.WithContext {
...
-          exit <- environment match {
-                    case Right(e) =>
-                      e.use { env =>
+        ParsedConfigs.parse[IO](cfg).value.flatMap {
+          case Right(parsed) =>
+            for {
+              _ <- logger.info("Initialising resources for Enrich job")
+              file = parsed.configFile
+              environment = for {
+                              blocker <- Blocker[IO]
```
chuwy, line 63: It seems like even more code could be extracted into a common module. Basically the only unique parts here are:

```
rawSource: Stream[F, Payload[F, Array[Byte]]]
mkGoodSink: Resource[F, AttributedByteSink[F]]
piiSink: ...
badSink: ...
```

Maybe I'm overly optimistic, but I hope that eventually all assets would differ only in these four arguments.

chuwy (approved):
Tried to concentrate on extraction itself and it looks great! Just a couple of nits.

benjben:
I integrated your feedback @chuwy (1 commit / comment). I'm much happier with how it looks now with as much as possible in `Environment`, as before

benjben:
The changes in https://github.com/snowplow/enrich/pull/474/commits/47e3818199a4bb25b66d17279d24659fdda61f98 are that checkpointing is now performed in a `Pipe`.

`Payload` has been removed (it was containing a tuple with the record and its associated `ack` function).
`Environment` now takes another type parameter `A` that will be `ConsumerRecord` for PubSub and `KinesisRecord` for Kinesis.

The logic is that the source now emits a `Stream[F, A]`, and the initial value `A` is passed along the transformations until the checkpointer at the end.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala
```diff
@@ -0,0 +1,152 @@
...
+      }
+    }
+
+  private def sendToSentry[F[_]: Sync](error: Throwable, sentry: Option[SentryClient]): F[Unit] =
+    sentry match {
+      case Some(client) =>
+        Sync[F].delay(client.sendException(error)) >> Logger[F].info("Sentry report has been sent")
+      case None =>
+        Sync[F].unit
+    }
```
benjben, line 151: @chuwy was there a specific reason why it was performed outside `F` in [previous implementation](https://github.com/snowplow/enrich/blob/master/modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Main.scala#L80-L87) ?

chuwy: Hard to say. I was just afraid that if something really bad happens - the CE runtime could be broken, so `IO`s wouldn't be executed anymore. But that's a fairly dumb assumption - I cannot explain why/how it would happen, so agree to leave it the way you did.

spenes (approved):
Looks really good!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala
```diff
@@ -0,0 +1,152 @@
...
+                      val env = Environment
+                        .make[F, Array[Byte]](
+                          blocker,
+                          ec,
+                          parsed,
+                          Source.filesystem[F](blocker, p.dir),
+                          goodSink,
+                          piiSink,
+                          badSink,
+                          _.flatMap(_ => Stream.eval(Sync[F].unit)),
```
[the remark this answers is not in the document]
benjben: Indeed, much better!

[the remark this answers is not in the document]
chuwy: Just `.void` even.

build.sbt
```diff
@@ -247,6 +242,29 @@ lazy val pubsub = project
...
+  .in(file("modules/pubsub"))
+  .dependsOn(commonFs2)
+  .settings(BuildSettings.basicSettings)
+  .settings(BuildSettings.formatting)
+  .settings(BuildSettings.scoverageSettings)
+  .settings(BuildSettings.sbtAssemblySettings)
+  .settings(
+    name := "snowplow-enrich-pubsub",
+    description := "High-performance streaming Snowplow Enrich job for PubSub built on top of functional streams",
+    buildInfoKeys := Seq[BuildInfoKey](organization, name, version, description),
```
chuwy, line 256: Nitpick: I think settings coming from separate plugins (`sbt-buildinfo` in this case should be in a separate list, like `sbtAssemblySettings` or `scoverageSettings`). Not a hard requirement though.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -69,35 +71,45 @@ object Enrich {
...
+  def run[F[_]: Concurrent: ContextShift: Clock: Parallel, A](env: Environment[F, A], ordered: Boolean): Stream[F, Unit] = {
     val registry: F[EnrichmentRegistry[F]] = env.enrichments.get.map(_.registry)
     val enrich: Enrich[F] = {
       implicit val rl: RegistryLookup[F] = env.registryLookup
-      enrichWith[F](registry, env.igluClient, env.sentry, env.metrics.enrichLatency)
+      enrichWith[F](registry, env.igluClient, env.sentry, env.metrics.enrichLatency, env.processor)
     }
 
+    val enrichPipe: Pipe[F, (A, Array[Byte]), (A, Result)] =
+      if(ordered)
```
chuwy, line 85: Shouldn't this formatting break the scalafmt check?

chuwy: It seems that for `ordered = true` we use `parEvalMapUnordered` while it should be other way round?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -69,35 +71,45 @@ object Enrich {
...
+        _.parEvalMapUnordered(ConcurrencyLevel){ case (orig, bytes) => enrich(bytes).map((orig, _)) }
+      else
+        _.parEvalMap(ConcurrencyLevel){ case (orig, bytes) => enrich(bytes).map((orig, _)) }
```
chuwy, line 88:
```suggestion
      in =>
        val parallel = if (ordered) in.parEvalMap(ConcurrencyLevel) else in.parEvalMapUnordered(ConcurrencyLevel) 
        parallel { case (orig, bytes) => enrich(bytes).map((orig, _)) }
```
 
Haven't checked if it compiles - you might need to help compiler to infer the type, but it highlights that `parEvalMapUnordered`/`parEvalMap` is the only difference.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -106,79 +118,84 @@ object Enrich {
         etlTstamp <- Clock[F].realTime(TimeUnit.MILLISECONDS).map(millis => new DateTime(millis))
         registry <- enrichRegistry
         enriched <- EtlPipeline.processEvents[F](adapterRegistry, registry, igluClient, processor, etlTstamp, payload)
-        trackLatency = enrichLatency(collectorTstamp)
-      } yield Payload(enriched, trackLatency *> row.finalise)
+        _ <- enrichLatency(collectorTstamp)
```
chuwy, line 121: Does it mean that we're changing the way we track latency? Previously it was guaranteed that the latency is tracked after the payload is back in the queue and acked. Now it seems it will be happening earlier - once we enriched the event?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -106,79 +118,84 @@ object Enrich {
...
           Failure.SizeViolation(Instant.now(), MaxRecordSize, size, msg),
           BadRowPayload.RawPayload(asStr.take(MaxErrorMessageSize))
         )
       Left(br)
     } else Right(asBytes)
   }
 
-  def sinkResult[F[_]: Concurrent: Parallel](env: Environment[F])(result: Validated[BadRow, EnrichedEvent]): F[Unit] =
+  def sinkAll[F[_]: Concurrent: Parallel, A, B](f: B => F[Unit]): Pipe[F, (A, List[B]), A] =
+    _.parEvalMap(SinkConcurrency){case (a, bs) => bs.parTraverse_(f).map(_ => a)}
```
chuwy, line 196:
```suggestion
    _.parEvalMap(SinkConcurrency) {case (a, bs) => bs.parTraverse_(f).as(a) }
```

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -192,4 +209,14 @@ object Enrich {
...
+  /**
+   * Controls the maximum number of events we can be waiting to get sunk
+   *
+   *  For the Pubsub sink this should at least exceed the number events we can sink within
+   *  `Sink.DefaultDelayThreshold`.
+   *
+   *  For the FileSystem source this is the primary way that we control the memory footprint of the
+   *  app.
+   */
+  private val SinkConcurrency = 10000
```
chuwy, line 221: Don't we want to make it configurable at least through `application.conf`?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala
```diff
@@ -0,0 +1,152 @@
...
+    environment: Resource[F, Environment[F, A]],
+    ordered: Boolean
+  ): F[ExitCode] =
+    environment.use { env =>
+      val log = Logger[F].info("Running enrichment stream")
+      val enrich = Enrich.run[F, A](env, ordered)
+      val updates = Assets.run[F, A](env)
+      val reporting = env.metrics.report
+      val flow = enrich.merge(updates).merge(reporting)
+      log >> flow.compile.drain.attempt.flatMap {
```
chuwy, line 135: I've been using `attempt.flatMap` pattern extensively until recently when I realised there are plenty of `recover`-like function that do exactly the same.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
```diff
@@ -0,0 +1,43 @@
...
+import com.snowplowanalytics.snowplow.enrich.common.fs2.ByteSink
+
+object Sink {
+
+  def fileSink[F[_]: Concurrent: ContextShift](path: Path, blocker: Blocker): Resource[F, ByteSink[F]] =
+    for {
+      channel <- Resource.fromAutoCloseableBlocking(blocker)(
+                   Sync[F].delay(FileChannel.open(path, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE))
+                 )
+      sem <- Resource.eval(Semaphore(1L))
```
chuwy, line 33: Can you explain in a comment why we need a lock here.

chuwy (changes requested):
Hey @benjben! I did the first pass. I think the only requested change at the moment is wrong `ordered` clause, but I also would like to hear about change of latency tracking.

Meanwhile I also would like to do one, more careful pass on sinks implementation

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: :man_facepalming: good catch !

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: You're right this is not correct, I need to move the latency to after the sinking.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
benjben: I left this part as it was before. My understanding is that several threads can share the same file sink and thus trying to write to the same file at the same time

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala
```diff
@@ -0,0 +1,157 @@
...
+                          processor,
+                          maxRecordSize
+                        )
+                      runEnvironment[F, A](env, ordered)
+                  }
+              } yield exit
+            }
+        ).flatten
+      case Left(error) =>
+        Sync[F].delay(System.err.println(error)) >> Sync[F].pure(ExitCode.Error)
```
chuwy, line 116: Same with logging.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -0,0 +1,67 @@
...
+
+  def mkHttp[F[_]: ConcurrentEffect](ec: ExecutionContext): Resource[F, HttpClient[F]] =
+    BlazeClientBuilder[F](ec).resource
+
+  case class DownloadingFailure(uri: URI) extends Throwable {
+    override def getMessage: String = s"Cannot download $uri"
+  }
+}
+
+trait Client[F[_]] {
```
chuwy, line 64: Can we move it inside `Clients` object at least? Just don't like having entities non matching with file name.

chuwy (approved):
LGTM!

Also, just in case `Http4sRegistryLookup.scala` is part of Iglu Scala Client 1.1.1

PR #476, opened 2021-07-16
Stream Enrich: rename directories (close #475)

### LINKED ISSUE #475: Stream Enrich: rename directories
At the moment in the repo we have `modules/kinesis`, `modules/kafka` and `modules/nsq` for Stream Enrich and we have `modules/pubsub` for PubSub with fs2.

The long term plan is to use fs2 for each queue and at some point deprecate Stream Enrich assets.

Stream Enrich assets should be renamed to `modules/stream-` (e.g. `modules/kinesis` -> `modules/stream-kinesis`) so that `modules/kinesis` would be the fs2 asset that we want long term

benjben:
I moved everything Stream Enrich related into `modules/stream/`


oguzhanunlu (approved):
LGTM 👍

PR #479, opened 2021-07-21
Base image update & Scio upgrade

oguzhanunlu:


project/Dependencies.scala
```diff
@@ -23,7 +23,9 @@ object Dependencies {
      // For Twitter's util functions
     "Twitter Maven Repo"             at "https://maven.twttr.com/",
     // For legacy Snowplow libs
-    ("Snowplow Analytics Maven repo" at "http://maven.snplow.com/releases/").withAllowInsecureProtocol(true)
+    ("Snowplow Analytics Maven repo" at "http://maven.snplow.com/releases/").withAllowInsecureProtocol(true),
```
istreeter, line 26: I know it's not connected to the current PR, but do we still need the Snowplow Maven repo?  Which libs are we pulling?

project/BuildSettings.scala
```diff
@@ -132,14 +132,11 @@ object BuildSettings {
   /** Docker settings, used by BE */
   lazy val dataflowDockerSettings = Seq(
     Docker / maintainer := "Snowplow Analytics Ltd. <support@snowplowanalytics.com>",
-    dockerBaseImage := "snowplow/k8s-dataflow:0.2.0",
-    dockerEnvVars := Map("HOME" -> "/tmp", "JAVA_OPTS" -> "-Duser.home=$HOME"),
-    Docker / daemonUser := "snowplow",
+    dockerBaseImage := "adoptopenjdk:11-jre-hotspot-focal",
+    Docker / defaultLinuxInstallLocation := "/opt/snowplow",
+    Docker / daemonUser := "daemon",
+    dockerEnvVars := Map("JAVA_OPTS" -> "-Duser.home=$HOME"),
```
istreeter, line 138: I'm very interested in this line!  Which dependency needs `user.home` to be set?  is it a dataflow thing?  And which directory does it use by default if this system property is not set?

It would be nice if `JAVA_OPTS` is kept as an environment variable that the user is allowed to change at run time, e.g.

```
docker run -e "JAVA_OPTS=-Dsome.other.property=something" snowplow/beam-enrich
```

But that example would now overwrite the `-Duser.home=$HOME` setting.

Having said that..... I am completely OK if there is no better way of making this work.

project/Dependencies.scala
```diff
@@ -23,7 +23,9 @@ object Dependencies {
...
+    // confluent repo
+    "confluent" at "https://packages.confluent.io/maven/"
```
istreeter, line 28: Could you leave a code comment explaining what we pull from the confluent repo?  Otherwise it looks a bit mysterious.

istreeter (commented):
Looks great @oguzhanunlu !  This will make maintenance much easier.

I left a couple of small questions.

project/Dependencies.scala
oguzhanunlu: probably not, let me check if we are fine without it

project/BuildSettings.scala
oguzhanunlu: Scio [uses](https://github.com/spotify/scio/blob/0539a4678d702503919f234be20a604cef4b13dc/scio-core/src/main/scala/com/spotify/scio/ScioContext.scala#L147) `user.home` java system property for an internal regex check.

When `user.home` is `?`, Scio fails with

```scala
Exception in thread "main" java.util.regex.PatternSyntaxException: Dangling meta character '?' near index 0  ?/\..+/.+
```

and `?` happens when there is no corresponding username for a given (UID, GID) within the container. That line was a dirty hack to make Scio think that `user.home` isn't `?`.

While writing this comment, I just noticed a simple mistake. The reason of this issue is not having [this configuration](https://github.com/snowplow/enrich/blob/master/project/BuildSettings.scala#L128) for beam enrich as well. Without `Docker / daemonUserUid := None`, beam-enrich isn't run with `daemon` but with `USER 1001`. Using `daemonUserUid ` config makes beam-enrich get executed by `daemon` user where `user.home` isn't `?`. I just pushed a new commit on top of this branch to be squashed after the review.

oguzhanunlu: Thank you for your interest @istreeter ! what looks like a small mistake (not having `daemonUserUid` config for beam-enrich) made me spend a good time diving into scio codebase to find the regex check followed by finding a workaround etc.

project/Dependencies.scala
oguzhanunlu: I can't run `assembly` or `docker:publishLocal` for `beam` project due to following errors

```scala
Error downloading io.confluent:kafka-avro-serializer:5.3.2
...
Error downloading io.confluent:kafka-schema-registry-client:5.3.2
...
```

What's is weird is that `sbt-dependency-graph` plugin's `dependencyList` doesn't list either of above artifacts for neither `beam` nor `common` project.

I added it to resolve the issue but I'll dig further to see where it comes from. It is also weird that somehow a kafka artifact is used to build beam-enrich. Maybe a plugin? I don't know yet.

project/Dependencies.scala
chuwy: It must be a Thrift-based `CollectorPayload`

oguzhanunlu: from coursier output

```
Downloading http://maven.snplow.com/releases/com/snowplowanalytics/snowplow-thrift-raw-event/0.1.0/
Downloading http://maven.snplow.com/releases/com/snowplowanalytics/schema-sniffer-1/0.0.0/
Downloading http://maven.snplow.com/releases/com/snowplowanalytics/collector-payload-1/0.0.0/
```

project/Dependencies.scala
oguzhanunlu: alright, coursier helped with this one,

```
beam (configurations compile, compile-internal, optional, provided, runtime, runtime-internal, test, test-internal)
└─ io.confluent:kafka-avro-serializer:5.3.2
   └─ org.apache.beam:beam-sdks-java-io-kafka:2.30.0
      └─ org.apache.beam:beam-runners-google-cloud-dataflow-java:2.30.0
```

Now I'll add a comment on why confluent registry is added

oguzhanunlu: I added the explanation in the github issue instead of the codebase and put a reference

oguzhanunlu:
I think it is ready now

istreeter (approved):

istreeter:
I assume this PR was not needed in the end, because Enrich 3.0.0 deprecated beam-enrich.  I will close it, but someone can re-open it if I am wrong.

PR #481, opened 2021-07-22
enrich-kinesis

benjben:
- [x] #505 
- [x] #506
- [x] #515 
- [x] #491 #492
- [x] #508 #509
- [x] #530
- [x] [fix initialTimestamp](https://github.com/snowplow-devops/terraform-modules/pull/2882#discussion_r769741854)
- [x] #487 Do we wait for permutive release ? => 1) test throughput with https://github.com/permutive/fs2-google-pubsub/pull/330 2) if not successful test with our own `0.18.0.1-M3`
- [x] #542
- [x] #517 #521 => :warning: some fields in `EnrichedEvent` don't match the type in `atomic` schema (e.g. [here](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L230-L232) and [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala#L124))
- [x] Add possibility to specify [maxConcurrency and bufferSize](https://github.com/laserdisc-io/fs2-aws/blob/main/fs2-aws-kinesis/src/main/scala/fs2/aws/kinesis/KinesisSettings.scala#L42-L43) and maybe others
- [x] #544 and add `% Runtime` to sts
- [x] make validation latest configurable
- [x] #531
- [ ] go over each change and think about what could break something


config/spirit-walk.hocon
```diff
@@ -0,0 +1,22 @@
+{
```
istreeter, line 1: Was this file committed by accident?

istreeter (commented):
I've only given this a quick look, but it looks very good indeed 🎉

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Source.scala
```diff
@@ -12,22 +12,32 @@
  */
 package com.snowplowanalytics.snowplow.enrich.common.fs2.io
```
voropaevp, line 13: `io` is shadowing namespace of dependencies like `io.circe`. This confuses IDE when importing `io.*` in `test`.
I suggest changing it to `inout` or some other name that is not used as a top-level domain.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
+      _
+        .evalTap(chunk => Logger[F].debug(s"${DateTime.now().toString()} - Starting to sink chunk of size ${chunk.size}"))
+        .parEvalMap(env.streamsSettings.concurrency.sink) { chunk =>
+          val records = chunk.map(_._1)
+          val results = chunk.map(_._2)
+          sinkChunk(results, sinkOne(env), env.metrics.enrichLatency).as(records)
+        }
+        .prefetch
+        .evalTap(chunk => Logger[F].debug(s"${DateTime.now().toString()} - Starting to checkpoint chunk of size ${chunk.size}"))
+        .evalMap(env.checkpoint)
```
istreeter, line 102: The old version of stream enrich checkpointed whenever [timeLimit](https://github.com/snowplow/enrich/blob/2.0.3/config/config.hocon.sample#L184) is reached.  There was no default, but in open source terraform modules [the default is 500 ms](https://github.com/snowplow-devops/terraform-aws-enrich-kinesis-ec2/blob/1ca899400ff159db2fbf4fc0c7fcc64ba2736881/variables.tf#L142).

Whereas the new enrich checkpoints after every batch, which by default is [10000 records](https://github.com/snowplow/enrich/blob/5778418d5ccf8ace73844cf5ed0110b768e3407c/modules/kinesis/src/main/resources/application.conf#L8).

I'm think I'm OK with this.  But I might as well tag @jbeemster who might have a comment on that!

modules/kinesis/src/main/resources/application.conf
```diff
@@ -0,0 +1,55 @@
+{
```
istreeter, line 1: How many collector payloads do you think it's possible to be holding in memory at the same time with these default settings?  If you're processing, say, 5 kinesis shards per enrich instance.

I count 190,000 (admittedly a worse case scenario).  And a further 10,000 more per additional shard.

Here's how:

Each polling batch is [10000 records](https://github.com/snowplow/enrich/blob/5778418d5ccf8ace73844cf5ed0110b768e3407c/modules/kinesis/src/main/resources/application.conf#L8).  The fs2-aws library uses a buffer that holds [10 batches](https://github.com/laserdisc-io/fs2-aws/blob/d07461a0d2515688bd4f18f87e8c14b80acd8773/fs2-aws/src/main/scala/fs2/aws/kinesis/KinesisSettings.scala#L43).  Plus there can be extra batch _per shard_ [on this line waiting to enqueue](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/Kinesis.scala#L86).

Then you have the batch you're currently processing, plus three extra batches because of the prefetches [here](https://github.com/snowplow/enrich/blob/5778418d5ccf8ace73844cf5ed0110b768e3407c/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L81), [here](https://github.com/snowplow/enrich/blob/5778418d5ccf8ace73844cf5ed0110b768e3407c/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L90) and [here](https://github.com/snowplow/enrich/blob/5778418d5ccf8ace73844cf5ed0110b768e3407c/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L100)

This is just to explain why I've been questioning memory problems, especially for large collector payloads.  Understanding the scope of this problem will be difficult, because we would need to explore regular processing, and also error-handling scenarios, e.g. kinesis problems, enrichment problems.  The latter are harder to predict or simulate.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
-      .pauseWhen(env.pauseEnrich)
-      .evalTap(_ => env.metrics.rawCount)
-      .fproduct(env.getPayload)
-      .through(enrichPipe)
-      .through(sinkResult(sinkOne(env), env.metrics.enrichLatency, env.streamsSettings.concurrency.output))
-      .through(env.checkpointer)
+    val enriched =
+      env.source.chunks
+        .prefetch
+        .evalTap(chunk => Logger[F].debug(s"${DateTime.now().toString()} - Starting to process chunk of size ${chunk.size}"))
```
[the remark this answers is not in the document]
benjben: I added the timestamp because it was not appearing on my local machine. It might get added by Cloudwatch and Stackdriver in the UI. Anyway I'm removing it

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
-    env.source
-      .pauseWhen(env.pauseEnrich)
-      .evalTap(_ => env.metrics.rawCount)
-      .fproduct(env.getPayload)
-      .through(enrichPipe)
-      .through(sinkResult(sinkOne(env), env.metrics.enrichLatency, env.streamsSettings.concurrency.output))
-      .through(env.checkpointer)
+    val enriched =
+      env.source.chunks
+        .prefetch
```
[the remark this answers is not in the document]
benjben: The idea of this `.prefetch` is that when the previous chunk is being enriched, next one is already getting ready and will be available for enriching as soon as the enriching is done.

You might be right that it does not make a difference. I'll try without it and see.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
+        .evalTap(chunk => Logger[F].debug(s"${DateTime.now().toString()} - Starting to process chunk of size ${chunk.size}"))
+        .pauseWhen(env.pauseEnrich)
```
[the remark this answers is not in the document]
benjben: Yes I remember our discussion I was planning to address it in the other PR, but I'll address it now.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
+        .evalTap(chunk => chunk.toList.map(_ => env.metrics.rawCount).sequence_)
```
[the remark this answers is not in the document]
benjben: Good point !

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -64,33 +68,40 @@ object Enrich {
...
+        .map(chunk => chunk.map(a => (a, env.getPayload(a))))
+        .evalTap(chunk => Logger[F].debug(s"${DateTime.now().toString()} - Starting to enrich chunk of size ${chunk.size}"))
```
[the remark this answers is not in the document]
benjben: I was using it with the timestamps to know where time was spent to find the bottleneck. Thanks to the size of the chunk it's possible to follow a chunk through the stages. Now that performance is fine we can probably remove it.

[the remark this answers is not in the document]
benjben: I wonder if there is value in keeping something like this :
```
.evalMap(chunk =>
          Clock[F].realTime(TimeUnit.MILLISECONDS).flatMap(before =>
            chunk.toList.map { case (orig, bytes) => enrich(bytes).map((orig, _)) }
              .parSequenceN(env.streamsSettings.concurrency.enrich).flatMap(results =>
                Clock[F].realTime(TimeUnit.MILLISECONDS).flatMap(after =>
                  Logger[F].info(s"Chunk of size ${chunk.size} enriched in ${after - before} ms")).as(results)
                )
          )
        )
```
Resulting in:
```
[pool-1-thread-2] INFO com.snowplowanalytics.snowplow.enrich.common.fs2.Enrich - Chunk of size 8974 enriched in 4114 ms
[pool-1-thread-2] INFO com.snowplowanalytics.snowplow.enrich.common.fs2.Enrich - Chunk of size 9195 enriched in 4740 ms
[pool-1-thread-1] INFO com.snowplowanalytics.snowplow.enrich.common.fs2.Enrich - Chunk of size 9205 enriched in 4350 ms
```
The code looks terrible but we could have a function for that. We could use it for enriching, sinking, checkpointing.

We might never need it though. It was just incredibly useful for me.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -135,63 +127,54 @@ object Environment {
...
+      good <- sinkGood
+      bad <- sinkBad
+      pii <- sinkPii.sequence
+      http <- Clients.mkHttp(ec)
+      clts = Clients.init[F](http, clients)
+      igluClient <- IgluClient.parseDefault[F](parsedConfigs.igluJson).resource
+      metrics <- Resource.eval(metricsReporter[F](blocker, file))
+      enrichments <- Enrichments.make[F](parsedConfigs.enrichmentConfigs, BlockerF.ofBlocker(blocker))
+      sem <- Resource.eval(Semaphore(1L))
+      _ <- Resource.eval(sem.acquire) // will be released at the end of assets initialization
```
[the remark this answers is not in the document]
benjben: You're right I've realized the same and forgot to move it!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: I've rewrote the assets mechanism so that [we use](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala#L80) a `Semaphore` instead of a `SignallingRef`.

Its permit is shared between [enrich](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L85) and [updating the assets](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L169).

The new logic for the refresh mechanism is to:
1. [Download all the assets](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L163) locally into a temporary location, without stopping the enriching
2. [Checks if some assets have changed](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L165) by comparing the hashes with the ones stored in the `Assets.State`
3. If there has been an update, then stop enriching and [update the assets](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L170)

There are 2 new unit tests :
1. [One](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/AssetsSpec.scala#L51-L85) that checks assets initialization
2. [One](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/AssetsSpec.scala#L110-L143) that inits the state with v1 of an asset, runs the update stream where v2 is now available, and checks that the asset has been updating to v2 locally

This can look weird that we are passing the [list of assets](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L156) to `updateStream` whereas we can get the information [from the enrichments](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala#L154). I had removed `List[Asset]` to then realize that with previous unit tests implementation it was not possible so I reverted the changes. Now that I rewrote the tests I think that we can make it work. Please let me know if you think that we should do it (not very important I think, just a neat), otherwise I'll leave it as is.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: `timeLimit` is not the checkpoint interval, but the maximum duration to which events are accumulated in a buffer before being sent to Kinesis. I wonder if checkpointing is triggered each time a buffer is sent to Kinesis.

With our design, we can't checkpoint more often, as we checkpoint records as soon as they have been sent to Kinesis.

We could find a way to checkpoint less often, but we are already checkpointing slower than once per minute, it should be totally fine with DynamoDB quotas.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +241,52 @@ object Enrich {
...
+            // The source has completed "naturally", e.g. processed all input files in the directory
+            Sync[F].unit
+          case (fiber, ExitCase.Canceled) =>
+            // SIGINT received. We wait for the enriched events already in the queue to get sunk and checkpointed
+            terminateStream(queue, fiber)
+          case (fiber, ExitCase.Error(e)) =>
+            // Runtime exception in the stream of enriched events.
+            // We wait for the enriched events already in the queue to get sunk and checkpointed.
+            // We then raise the original exception
+            terminateStream(queue, fiber) *> Sync[F].raiseError(e)
```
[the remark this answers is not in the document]
benjben: I get it now ! Thank you

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -152,31 +168,48 @@ object Enrich {
...
 
+  def sinkChunk[F[_]: Concurrent: Parallel, B](
+    results: List[Result],
+    sinkOne: Validated[BadRow, EnrichedEvent] => F[Unit],
+    trackLatency: Option[Long] => F[Unit]
+  ): F[Unit] =
+    results.parTraverse_ {
+      case (events, collectorTstamp) =>
+        events.parTraverse_(one => sinkOne(one)) <* trackLatency(collectorTstamp)
+    }
```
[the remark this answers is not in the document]
benjben: That's a good point. Unfortunately we can't have such `.flush` function as all we have access to on the underlying producer is [putData](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/internal/KinesisProducerClient.scala#L15).

What we can do however is to use `2` as `concurrency.sink` here https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L93 so that next chunk starts getting sunk already when the previous one finishes (and is possibly in the 100ms interval). I'm going to try to see the stats.

`fs2-aws` doesn't add much on top of the underlying `KinesisProducer`, we could also discard it and use our own implementation directly (for producing).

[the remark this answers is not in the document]
benjben: Interestingly, the best throughput is observed with `sink.concurrency` set to 2 (= number of cores) and without any `prefetch` (adding `prefetch` has no impact).

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: See https://github.com/snowplow/enrich/pull/481#discussion_r768879269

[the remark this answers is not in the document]
benjben: Alright I've looked at [here](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/Kinesis.scala#L94) and I totally agree that this `.prefetch` was never useful for Kinesis.

modules/kinesis/src/main/resources/application.conf
benjben: Very interesting!

I agree with your way of counting. There has been only a little change, I removed the `.prefetch` to now use `>1` for `sink.concurrency` [here](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L93) so now it's `concurrency - 1` additional shards instead of three.

Because of the enqueuing for each shard I'm afraid the memory used by the app is unbounded. So we need to make sure that our auto-scaling is good enough so that new instances are spawned before the app gets a chance to go OutOfMemory. I'll look at our scaling strategy in Terraform. But I think that whatever the strategy, it would be great to use the average record size of a customer to determine the memory allocated to the JVM. /cc @oguzhanunlu @jbeemster

benjben:
Commits worth reviewing since the latest review:
1. https://github.com/snowplow/enrich/pull/481/commits/daca7adb10f4d814621bbc01491f569c4b6f3ca4
  More parameters are configurable for Kinesis sink and source.
  We now use chunks in [enrich workflow](modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala).
  [checkpoint](https://github.com/snowplow/enrich/commit/daca7adb10f4d814621bbc01491f569c4b6f3ca4#diff-88a301cb49a839af9f152365d1db31ee9b8a6c40ec52d20ba27a6b548596cf45R79-R93) function.
2. https://github.com/snowplow/enrich/pull/481/commits/37bb18c9e459b17243e6e61f0619ca438211ac1c
3. https://github.com/snowplow/enrich/pull/481/commits/33a8692a2029e312d74640035cd942799656f74d
  FYI. All the tests from Beam Enrich have been moved to common-fs2. [This](modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/blackbox/BlackBoxTesting.scala) is the most interesting part. 
4. https://github.com/snowplow/enrich/pull/481/commits/b3da4f6d175dc77da6bf61572d4d7fa7cc17f3c9
5. https://github.com/snowplow/enrich/pull/481/commits/7a99af0946551cc79345706e00f1f003ab928f75
6. https://github.com/snowplow/enrich/pull/481/commits/170b5437983634908cba59a7ee11f7820b067f7c
  This one is important as some fields of the [EnrichedEvent](modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala) have been updated because there was a discrepancy between some fields' type in `EnrichedEvent` and in `atomic` schema (e.g. [here](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0#L230-L232) and [here](https://github.com/snowplow/enrich/blob/master/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala#L124)) and therefore some tests were failing due to the new validation.
  This also introduces feature flags, so that the validation can be deactived for backward compatibility.
  I took the opportunity to put all the schemas used in the tests of `common-fs2` in `resources/iglu-client-embedded/`.
  I needed to rewrite [CurrencyConversionEnrichmentSpec](modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/CurrencyConversionEnrichmentSpec.scala) without specs2 data tables because there were errors impossible to spot.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
```diff
@@ -216,4 +217,21 @@ final case class CurrencyConversionEnrichment[F[_]: Monad](
         )
         Monad[F].pure(FailureDetails.EnrichmentFailure(enrichmentInfo, f).invalidNel)
     }
+
+  // The original error is either a Throwable or an OER error
+  private def mkEnrichmentFailure(error: Either[Throwable, OerResponseError]): FailureDetails.EnrichmentFailure = {
+    val msg = error match {
+      case Left(t) =>
+        val sw = new StringWriter
+        t.printStackTrace(new PrintWriter(sw))
```
istreeter, line 226: Please can you explain the justification for putting the entire stack trace into the bad row?  Isn't it quite unprecedented to do this?  I mean, there must be loads of places in our estate where we catch exceptions to create bad rows, but I have never seen a stack trace get serialized before.

I'm not necessarily against it, I just wondered if this is part of a wider plan.... and how the stacktrace might get used downstream.

What are our alternatives if we don't serialize the stack trace?

istreeter (commented):
There's so many things in this PR that I love :)

I haven't finished reviewing it yet, so I'm letting you have this first round of suggestions early.

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
```diff
@@ -0,0 +1,96 @@
...
+  private def getPayload(record: CommittableRecord): Array[Byte] = {
+    val data = record.record.data
+    val buffer = ArrayBuffer[Byte]()
+    while (data.hasRemaining())
+      buffer.append(data.get)
+    buffer.toArray
+  }
+
+  /** For each shard, the record with the biggest sequence number is found, and checkpointed. */
+  private def checkpoint[F[_]: Parallel: Sync](records: List[CommittableRecord]): F[Unit] =
```
istreeter, line 83: This is interesting, because the enrich batch size setting directly affects how often we checkpoint.  If you decrease batch size then you checkpoint more often; therefore might need to bump dynamodb write capacity to compensate.

Probably not a problem for 3.0.0, but definitely something for pipeline operators to be aware of.  Especially as that's not how stream-enrich worked.

config/config.kinesis.extended.hocon
```diff
@@ -0,0 +1,396 @@
...
+      # "cloudwatchPort": 4582
+    }
+  }
+
+  # Optional. Concurrency of the app
+  "concurrency" : {
+    # Number of events that can get enriched at the same time within a chunk
+    "enrich": 256
+    # Number of chunks that can get sunk at the same time
+    "sink": 1
```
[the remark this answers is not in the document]
benjben: Good point!

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
benjben: The reason is that it happened to me many times while troubleshooting the unit tests that the bad row did not contain enough information to find out what the issue was. Each time it happened I needed to update the code to print the stacktrace to identify the faulty line.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +232,50 @@ object Enrich {
...
+            // The source has completed "naturally", e.g. processed all input files in the directory
+            Sync[F].unit
+          case (fiber, ExitCase.Canceled) =>
+            // SIGINT received. We wait for the enriched events already in the queue to get sunk and checkpointed
+            terminateStream(queue, fiber)
+          case (fiber, ExitCase.Error(e)) =>
+            // Runtime exception in the stream of enriched events.
+            // We wait for the enriched events already in the queue to get sunk and checkpointed.
+            // We then raise the original exception
+            terminateStream(queue, fiber).attempt *> Sync[F].raiseError(e)
```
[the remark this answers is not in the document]
benjben: Great point !

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/blackbox/BlackBoxTesting.scala
```diff
@@ -0,0 +1,135 @@
...
+    expected: Map[String, String],
+    enrichmentConfig: Option[Json] = None
+  ) =
+    Enrich
+      .enrichWith(getEnrichmentRegistry(enrichmentConfig), igluClient, None, EnrichSpec.processor, true)(
+        input
+      )
+      .map {
+        case (List(Validated.Valid(enriched)), _) => checkEnriched(enriched, expected)
+        case other => ko(s"there should be one enriched event but got $other")
```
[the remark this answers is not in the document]
benjben: Agreed that it would be great but these black box tests contain all the tests migrated from Beam Enrich and there is no such case, so leaving it as is for now.
But don't worry, this use case of a `CollectorPayload` holding multiple events is tested in other places.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +232,50 @@ object Enrich {
...
+        }
+    }
+
+  private def terminateStream[F[_]: Concurrent: Sync: Timer, A](queue: NoneTerminatedQueue[F, A], fiber: Fiber[F, Unit]): F[Unit] =
+    for {
+      _ <- Sync[F].delay(println("Terminating enrich stream")) // can't use logger here because it might have been shut down already
```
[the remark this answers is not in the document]
benjben: Great point !

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
benjben: > batch size setting directly affects how often we checkpoint

I'm not sure to follow, which batch size are you talking about, `maxRecords`?

> might need to bump dynamodb write capacity to compensate

Well I think that if enrich goes beyond DynamoDB write capacity, it really means that enrich is misconfigured.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
istreeter: The max allowed length of that field is https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.badrows/enrichment_failures/jsonschema/2-0-0#L78

Can the strack trace exceed 512 characters?

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
istreeter: I meant `input.retrievalMode.maxRecords`.

Let's imagine enrich is processing one shard and can process 1000 events per second.

If `input.retrievalMode.maxRecords` is 10000, then enrich will checkpoint once every 10 seconds.  If `input.retrievalMode.maxRecords` is 100, then enrich will checkpoint ten times every seconds.  That might surprise someone that retrievalMode settings can influence checkpoint frequency, unless that person understands the internals of our code.

It's fine for you to resolve this conversation, and not make any changes.  I just think we should remember to document somewhere that `input.retrievalMode.maxRecords` has this slightly surprising consequence on something unrelated to retrieving records.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: Confirmed that the log appears

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +232,50 @@ object Enrich {
...
+      _ <- queue.enqueue1(None)
+      _ <- fiber.join.timeoutTo(10.seconds, fiber.cancel)
```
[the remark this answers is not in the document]
benjben: Great point!

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
benjben: Good catch ! Indeed it probably can. Removing it then

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
benjben: > has this slightly surprising consequence on something unrelated to retrieving records
I'm adding a note in the config

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
benjben: Actually I'll leave it inside and take a substring of the first 512 characters, if that's fine with you

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +232,51 @@ object Enrich {
...
+            // Runtime exception in the stream of enriched events.
+            // We wait for the enriched events already in the queue to get sunk and checkpointed.
+            // We then raise the original exception
+            terminateStream(queue, fiber).attempt *> Logger[F].error(e)("Unexpected error in enrich") *> Sync[F].raiseError(e)
+        }
+    }
+
+  private def terminateStream[F[_]: Concurrent: Sync: Timer, A](queue: NoneTerminatedQueue[F, A], fiber: Fiber[F, Unit]): F[Unit] =
+    for {
+      timeout <- Sync[F].pure(5.minutes)
```
istreeter, line 277: Why was this 5 minute timeout needed?  I was expecting termination to work smoothly without any explicit timeout like this.

istreeter: Oh - ignore this comment.  I mis-read it as sleeping for 5 minutes.  This is fine.

config/config.kinesis.extended.hocon
```diff
@@ -0,0 +1,396 @@
...
+    "retrievalMode": {
+      "type": "Polling"
+      "maxRecords": 10000
+    }
+    # "retrievalMode": {
+    #   "type": "FanOut"
+    # }
+
+    # Optional. Max size of the KinesisAsyncClient HTTP thread pool
+    "maxConcurrency": 100
```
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

config/config.kinesis.extended.hocon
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

config/config.kinesis.extended.hocon
```diff
@@ -0,0 +1,396 @@
...
+    instanceId = 665bhft5u6udjf
+
+    # Name of the terraform module that deployed the app
+    moduleName = enrich-kinesis-ce
+
+    # Version of the terraform module that deployed the app
+    moduleVersion = 1.0.0
+  }
+
+  # Optional. To activate/deactive enrich features
```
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/2f091f46fa6c22db60b3aca47f5e584a5727c6f4

config/config.kinesis.extended.hocon
```diff
@@ -0,0 +1,396 @@
...
+  "featureFlags" : {
+    # If activated, enriched events will get validated against atomic schema.
+    # If not valid, a bad row will be emitted instead of the enriched event.
+    "validateEnrichedEvents": true
```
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/2f091f46fa6c22db60b3aca47f5e584a5727c6f4

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
```diff
@@ -0,0 +1,96 @@
...
+    records
+      .groupBy(_.shardId)
+      .foldLeft(List.empty[CommittableRecord]) { (acc, shardRecords) =>
```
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CurrencyConversionEnrichment.scala
benjben: https://github.com/snowplow/enrich/pull/481/commits/be494ac19c4461d4b16307d8a8e44f104cb30ed8

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala
benjben: https://github.com/snowplow/enrich/pull/481/commits/57f72c1aae548e39f2dd540727f9ca3d5821cfa9

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/DbExecutor.scala
```diff
@@ -83,24 +83,20 @@ object DbExecutor {
...
-                        }
-        } yield connection
+          existingConn <- connectionRef.get(())
+          conn <- flattenCached(existingConn) match {
+                    case Right(conn) if !conn.isClosed =>
+                      Sync[F].pure(conn.asRight)
+                    // Connection closed or unitialized
+                    case Right(_) | Left(Unitialized) =>
+                      for {
+                        newConn <- Sync[F].delay(Either.catchNonFatal(DriverManager.getConnection(rdbms.connectionString)))
```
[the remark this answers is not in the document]
benjben: https://github.com/snowplow/enrich/pull/481/commits/7293107af16d15bcd06d2f7a441d474a3cc04c70
I used [HikariCP](https://github.com/brettwooldridge/HikariCP)

To preserve the old behavior and not crash the app if for instance SQL credentials are wrong, I couldn't instantiate `HikariDataSource` already when creating `SqlQueryEnrichment` because it can throw a `SQLException`.

I considered going with a `Ref[F, Option[HikariDataSource]]` initialized to `None` but we can't use `Ref` with `Id` (it's missing `Sync`).

So I [used](https://github.com/snowplow/enrich/pull/481/commits/7293107af16d15bcd06d2f7a441d474a3cc04c70#diff-800f2303e20ced7e5977daa3b33212e2fb927cb94c7b1271fbf985ea315e10f6R121) a `lazy val` so that it's accessed for the first time only when enriching an event, and creating a bad row in case of problem.

benjben:
Hey @istreeter , I addressed all your feedback. I added the commit holding the changes for each of your comment. I think that the most important one to re-review is [this one](https://github.com/snowplow/enrich/pull/481#discussion_r798434482).

benjben:
Hey @istreeter, I'm done with the changes re the validation with the feature flag and SQL enrichment.

For the feature flag, the validation now *always* happens, and depending on the feature flag we don't emit a bad row in case of invalid and we increment the new metric and log the bad row. This happens [here](https://github.com/snowplow/enrich/pull/481/commits/7c16debb9e6d48e4491754da4707468172b0ed98#diff-ffa77f6f8575319753246e40b59e71400506eb914050c569e78bb44987e47f90R816-R826).

For SQL enrichment, we need to explicitly give the `Connection` back to the pool (of size 1) when we are done with it. This is what `Connection.close()` does (it doesn't actually close the connection).
So the new logic is to get the connection [at the beginning](https://github.com/snowplow/enrich/pull/481/commits/c096910fd1ecf73bbee27d6aa666ac0d1c1a2e42#diff-800f2303e20ced7e5977daa3b33212e2fb927cb94c7b1271fbf985ea315e10f6R146) of the enrichment, to use it everywhere (statement is tied to the connection), and to "[close](https://github.com/snowplow/enrich/pull/481/commits/c096910fd1ecf73bbee27d6aa666ac0d1c1a2e42#diff-800f2303e20ced7e5977daa3b33212e2fb927cb94c7b1271fbf985ea315e10f6R229)" it [at the end](https://github.com/snowplow/enrich/pull/481/commits/c096910fd1ecf73bbee27d6aa666ac0d1c1a2e42#diff-800f2303e20ced7e5977daa3b33212e2fb927cb94c7b1271fbf985ea315e10f6R170).

The best way to do that would be to use `cats.effect.Ressource`, but because of `Id` we can't, so instead after each `EitherT` (a `Left` short-circuits the workflow), I manually close the connection in case of error (e.g. [here](https://github.com/snowplow/enrich/pull/481/commits/c096910fd1ecf73bbee27d6aa666ac0d1c1a2e42#diff-800f2303e20ced7e5977daa3b33212e2fb927cb94c7b1271fbf985ea315e10f6R157)). I'm aware that it doesn't look super pretty, but it works. Let me know if you think that it's better to create our own `ResourceF`.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/DbExecutor.scala
```diff
@@ -81,26 +83,8 @@ object DbExecutor {
...
-                                                         Either.catchNonFatal(DriverManager.getConnection(rdbms.connectionString))
-                                                       )
-                                            _ <- connectionRef.put((), newConn)
-                                          } yield newConn
-                            } yield result
-                          case Left(error) =>
-                            error.asLeft[Connection].pure[F]
-                        }
-        } yield connection
+      def getConnection(dataSource: HikariDataSource): F[Either[Throwable, Connection]] =
```
istreeter, line 86: I can see you are being very careful to catch every possible exception, and close the connection if anything goes wrong.  However, it might be easier to change the `DbExecutor` trait so that instead of `getConnection` it has `withConnection`.  This can ensure that the exception is always closed, and then you don't need to be so careful about it in other places in the code.

```
implicit def syncDbExecutor[F[_]: Sync]: DbExecutor[F] =
  new DbExecutor[F] {
    def withConnection[A](dataSource: HikariDataSource)(f: Connection => F[Either[Throwable, A]]): F[Either[Throwable, A]] =
      Sync[F].bracket(Sync[F].delay(dataSource.getConnection()))(f)(cx => Sync[F].delay(cx.close())
        .handleError(e => Left(e))
```

```
implicit def idDbExecutor: DbExecutor[Id] =
  new DbExecutor[Id] {
    def  withConnection[A](dataSource: HikariDataSource)(f: Connection => Either[Throwable, A]): Either[Throwable, A] =
      try {
        // get connection and use connection
      } finally {
        // close connection
     }
```

Just an idea.

benjben: Please see https://github.com/snowplow/enrich/pull/481/commits/64cc9585c2c7f3d145776af31be12e43fb71f821

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/DbExecutor.scala
```diff
@@ -32,8 +34,8 @@ import scala.collection.immutable.IntMap
 /** Side-effecting ability to connect to database */
 trait DbExecutor[F[_]] {
 
-  /** Lookup a connection from mutable reference or initialize and put into the cache */
-  def getConnection(db: Rdbms, connRef: ConnectionRef[F])(implicit M: Monad[F]): F[Either[Throwable, Connection]]
+  /** Get a connection from the Hikari data source */
+  def getConnection(dataSource: HikariDataSource): F[Either[Throwable, Connection]]
```
pondzix, line 38: I can see we rely here mainly on `java.sql`, so maybe we could use `javax.sql.DataSource` here and pass hikari as implementation?

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala
```diff
@@ -101,23 +107,25 @@ object SqlQueryEnrichment extends ParseableEnrichment {
...
   cache: SqlCache[F],
-  connection: ConnectionRef[F],
   blocker: BlockerF[F]
 ) extends Enrichment {
   private val enrichmentInfo =
     FailureDetails.EnrichmentInformation(schemaKey, "sql-query").some
 
+  private val logger = LoggerFactory.getLogger(getClass)
+
+  private val dataSource = getDataSource(db)
```
pondzix, line 127: Previously `ConnectionRef` was created in `CreateSqlQueryEnrichment`. Is there any reason `dataSource` is created here as `private val` instead of passing as constructor arg as before?

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala
```diff
@@ -134,18 +142,35 @@ final case class SqlQueryEnrichment[F[_]: Monad: DbExecutor](
     customContexts: List[SelfDescribingData[Json]],
     unstructEvent: Option[SelfDescribingData[Json]]
   ): F[ValidatedNel[FailureDetails.EnrichmentFailure, List[SelfDescribingData[Json]]]] = {
+    def useConnection(connection: Connection): F[Either[NonEmptyList[String], List[SelfDescribingData[Json]]]] =
```
pondzix, line 145: It looks like nested method, what do you think about converting it to simple `private def` in the class? IMO it could improve readability.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala
```diff
@@ -134,18 +142,35 @@ final case class SqlQueryEnrichment[F[_]: Monad: DbExecutor](
...
+                          .toEitherT[F]
+                          .leftMap { t =>
+                            NonEmptyList.of("Error while building the map of placeholders", t.toString)
+                          }
+        verifiedPlaceholders <-
+          EitherT(blocker.blockOn(Monad[F].pure(DbExecutor.allPlaceholdersFilled(connection, query.sql, placeholders))))
+            .leftMap { t =>
+              NonEmptyList.of("Error while filling the placeholders", t.toString)
+            }
+        result <- verifiedPlaceholders match {
```
pondzix, line 158: These steps in the forcomp are getting bigger, maybe they could be extracted to separate methods? I think it would be easier to follow what `useConnection` does :)

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala
```diff
@@ -0,0 +1,39 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the Apache License Version 2.0 for the specific language governing permissions and limitations there under.
+ */
+package com.snowplowanalytics.snowplow.enrich.common.utils
+
+import cats.Id
+
+import cats.effect.{Resource, Sync}
+
+trait ResourceF[F[_]] {
```
pondzix, line 19: This is very cool! 😎

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala
```diff
@@ -0,0 +1,39 @@
...
+  def use[A, B](acquire: A)(release: A => Unit)(use: A => F[B]): F[B]
```
pondzix, line 20: Wondering about `acquire` name as the resource is already 'acquired' when calling this method. Maybe just `resource`?

benjben:
@istreeter most of last week's changes are about Kinesis sink:
- [addition of `streamExists` to crash the app at startup if the stream doesn't exist (same for Kinesis source)](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L54)
- [Use maximum `recordTtl` so that KPL retries "forever"](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L100) 
- [Better logging in case of KPL failure](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L148-L154)
-  [Better logging in case of KPL error](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L157-L162)
- [Remove `raiseError` so that cats-retry also retries forever](https://github.com/snowplow/enrich/blob/feature/enrich-kinesis/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L163)

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala
benjben: What would be the benefit of doing so ?
I tried to simplify the old implementation. `dataSource` is never accessed anywhere else than inside `SqlQueryEnrichment`, and doesn't need to be initialized before.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala
benjben: Thanks ! Glad that you like that. Dealing with `Id` is a pain !

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -219,4 +242,51 @@ object Enrich {
...
+            // The source has completed "naturally", e.g. processed all input files in the directory
+            Sync[F].unit
+          case (fiber, ExitCase.Canceled) =>
+            // SIGINT received. We wait for the enriched events already in the queue to get sunk and checkpointed
+            terminateStream(queue, fiber)
+          case (fiber, ExitCase.Error(e)) =>
+            // Runtime exception in the stream of enriched events.
+            // We wait for the enriched events already in the queue to get sunk and checkpointed.
+            // We then raise the original exception
+            terminateStream(queue, fiber).attempt *> Logger[F].error(e)("Unexpected error in enrich") *> Sync[F].raiseError(e)
```
istreeter, line 281: Hey @benjben , I was lucky enough to have a pipeline that threw an exception!  So I was able to see what logs we get on error:

```
[pool-1-thread-4] WARN com.snowplowanalytics.snowplow.enrich.common.fs2.Enrich - Terminating enrich stream. Waiting 5 minutes for it to complete                                                                                            
[pool-1-thread-8] ERROR com.snowplowanalytics.snowplow.enrich.common.fs2.Run - The Enrich job has stopped                                                                                                                                   
[pool-1-thread-8] INFO org.http4s.client.PoolManager - Shutting down connection pool: curAllocated=1 idleQueues.size=1 waitQueue.size=0 maxWaitQueueLimit=256 closed=false                                                                  
java.lang.NullPointerException                                                                                                                                                                                                              
        at io.circe.Printer$PrintingFolder.onString(Printer.scala:309)                                                                                                                                                                      
        at io.circe.Printer$PrintingFolder.onString(Printer.scala:292)   
  -- etc
  ```

Do you think it's a possible problem that it does not print the error message until after it has shut down the stream?  I was thinking if the orchestration (e.g. docker/kubernetes) sends a forceful kill signal while Enrich is still trying to terminate, then you never see the log message at all.

In BigQuery loader I made it so that it [logs the error message before trying to terminate the stream](https://github.com/snowplow-incubator/snowplow-bigquery-loader/pull/250/files#diff-a92f3fcd4f8171febee7c6990a044ba161df3cb662414046bebe3772e7dabbaeR111).  So if something goes wrong after that, then at least we know the error message has been printed.

It's a small detail, sorry for nitpicking.

Comment (benjben) [reply modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala:281]: Yeah good point, I'll put the log before.

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala (hunk 28 above)
Comment (pondzix) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala:127]: To me `CreateSqlQueryEnrichment` seems like a place where we create/init necessary dependencies. Then, after 'creator' does his job, we can use those injected dependencies in our enrichment class and not think about initialising anything. With `dataSource` created here it can be a bit confusing like `oh, here is one more thing we need to the job` 🙂 But it's really a nit and can remain as it is. 

BTW do we need separate those 2 instances of `CreateSqlQueryEnrichment` for sync and Id? Can't we rely on `F[_]`? It looks like the body of those implementations is the same.

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala (hunk 31 above)
Comment (pondzix) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala:19]: Yeah, I can see `Id` is basically everywhere. Out of curiosity: why we need to support `Id` in enrich? I know about rdb loader/shredder and spark thing, but enrich?

Comment (chuwy) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/ResourceF.scala:19]: Hopefully it's going to be a past very soon.

Current Stream Enrich is built on KCL, without anything from Cats Effect / FS2. Also Beam Enrich isn't that distant past (we've just EOL'ed it few months ago). Not to mention we had Spark Enrich at some point.

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala (hunk 28 above)
Comment (benjben) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/SqlQueryEnrichment.scala:127]: Yeah you're right, I'll move it to `CreateSqlEnrichment` :+1: 

> It looks like the body of those implementations is the same.

And again you're right !

File: modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/package.scala [hunk 34]
```diff
@@ -0,0 +1,42 @@
...
+  def getRuntimeRegion: Option[String] =
+    Option((new DefaultAwsRegionProviderChain).getRegion)
+
+  def streamExists[F[_]: Sync: ContextShift](region: String, streamName: String): F[Unit] =
+    Resource
+      .fromAutoCloseable(
+        Sync[F].delay(KinesisClient.builder().region(Region.of(region)).build)
+      )
+      .use { client =>
+        val describeStreamRequest = DescribeStreamRequest
```
benjben: From @istreeter : in collector we made it so that startup checks like that are disabled.  This was actually really important.... there is a account-wide rate limit on the kinesis describeStream endpoint.  If there's a large scaling event then we can easily exceed the rate limit and it fails.

PR #484, opened 2021-08-02
Enrich 3.0.0 : bunch of smaller changes

istreeter:
We intended to do this for the 2.0.0 release, but at the time I couldn't work out how to use a `HotSwap`.  I worked on this issue because I wanted to learn how to do it properly.

For backwards compatibility.... if `maxBytes` is not set then it creates a single output file - just like before.  But if `maxBytes` is defined then it creates many files and adds a `.0001`, `.0002` etc suffix.


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
```diff
@@ -40,4 +48,79 @@ object Sink {
...
+    Resource.fromAutoCloseableBlocking(blocker) {
+      Sync[F].delay(FileChannel.open(path, StandardOpenOption.CREATE_NEW, StandardOpenOption.WRITE))
+    }
+
+  private def makeOpenState[F[_]: Sync: ContextShift](
+    blocker: Blocker,
+    index: Int,
+    base: Path
+  ): Resource[F, OpenState] = {
+    val path = base.resolveSibling(f"${base.getFileName}%s.$index%04d")
```
benjben, line 92: Neat!

benjben: What happens if enrich runs forever and we have > 9999 files ?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
```diff
@@ -40,4 +48,79 @@ object Sink {
...
+   *  Each file has an integer suffix e.g. /path/to/good.0001
+   */
+  def rotatingFileSink[F[_]: Concurrent: ContextShift](
+    path: Path,
+    maxBytes: Long,
+    blocker: Blocker
+  ): Resource[F, ByteSink[F]] =
+    for {
+      hs <- Hotswap.create[F, OpenState]
+      ref <- Resource.eval(Ref.of[F, RotatingFileState](ClosedState(1)))
```
benjben, line 61: Can't we start directly with an `OpenState` (that would then become `FileState`) to get rid of this resource and have only the `HotSwap` ?

benjben (approved):
Awesome @istreeter !

Could be nice to add a unit test.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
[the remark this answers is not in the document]
istreeter: The 10000th file has suffix `.10000`.  So it breaks alphabetical ordering of the file names, but apart from that it continues to work OK.

```
scala> f"basename.${1}%04d"
res0: String = basename.0001

scala> f"basename.${9999}%04d"
res1: String = basename.9999

scala> f"basename.${10000}%04d"
res2: String = basename.10000
```

config/config.kinesis.reference.hocon
```diff
@@ -202,4 +204,34 @@
       "cloudwatch": false
     }
   }
+
+  # Optional, configure telemetry
+  # All the fields are optional
+  telemetry {
```
chuwy, line 210: Seems like non-JSONy style, unaligned with rest of the file.

config/config.kinesis.reference.hocon
```diff
@@ -202,4 +204,34 @@
...
+    # Set to true to disable telemetry
+    disable = false
+    # Interval for the heartbeat event
+    interval = 15 minutes
+    # HTTP method used to send the heartbeat event
+    method = POST
+    # URI of the collector receiving the heartbeat event
+    url = collector-g.snowplowanalytics.com
```
chuwy, line 218: Also not consistent how we configure collector in other places (`collectorUri` is used I think?)

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Sink.scala
```diff
@@ -40,4 +48,75 @@ object Sink {
...
 
+  /**
+   * Opens a new file when the existing file exceeds `maxBytes`
+   *  Each file has an integer suffix e.g. /path/to/good.0001
+   */
+  def rotatingFileSink[F[_]: Concurrent: ContextShift](
+    path: Path,
+    maxBytes: Long,
+    blocker: Blocker
+  ): Resource[F, ByteSink[F]] =
```
chuwy, line 59: FYI, `Resource` can be safely flatMap-ped into plain `Pipe`

project/Dependencies.scala
```diff
@@ -43,27 +43,27 @@ object Dependencies {
 
     val refererParser    = "1.1.0"
     val maxmindIplookups = "0.7.1"
-    val circe            = "0.13.0"
-    val circeOptics      = "0.13.0"
+    val circe            = "0.14.0"
```
chuwy, line 46: 0.14.1 is available.

project/Dependencies.scala
```diff
@@ -43,27 +43,27 @@ object Dependencies {
...
     val circeConfig      = "0.7.0"
-    val circeJackson     = "0.13.0"
+    val circeJackson     = "0.14.0"
     val scalaForex       = "1.0.0"
     val scalaWeather     = "1.0.0"
     val gatlingJsonpath  = "0.6.14"
     val scalaUri         = "1.4.5"
-    val badRows          = "2.1.0"
+    val badRows          = "2.1.1"
     val igluClient       = "1.0.2"
```
chuwy, line 55: 1.1.1 is available.

chuwy (approved):
Looks good. Just couple more libs can be bumped.

Issue #487, opened 2021-08-11
Pubsub: use custom fs2-google-pubsub-grpc dependency


oguzhanunlu:
Our custom version lives [here](https://github.com/snowplow-incubator/fs2-google-pubsub/tree/release/0.18.0)

Ours is rebased on original 0.17.0 tag, I didn't include further updates from original project since the library was migrated to cats effect 3.

benjben:
Not needed any more as permutive integrated our changes in their latest release

PR #488, opened 2021-08-11
Pubsub: use custom fs2-google-pubsub-grpc dependency

oguzhanunlu:


project/Dependencies.scala
```diff
@@ -69,7 +69,7 @@ object Dependencies {
     val decline          = "1.0.0"
     val fs2              = "2.5.5"
     val catsEffect       = "2.5.0"
-    val fs2PubSub        = "0.16.1"
+    val fs2PubSub        = "0.18.0-M8"
```
benjben, line 72: Why don't we release final version?

benjben (approved):
:+1:

project/Dependencies.scala
oguzhanunlu: just addressed

benjben:
See https://github.com/snowplow/enrich/issues/487#issuecomment-1021012957

PR #514, opened 2021-11-08
Add Adapted type

chuwy:
Just been experimenting over the weekend with my idea of `snowplow-core` idea and realised we have this horrible type repeated over and over: `ValidatedNel[FailureDetails.AdapterFailureOrTrackerProtocolViolation, NonEmptyList[RawEvent]]`, which in fact is quite central to the enrich.

@benjben, feel free to cherry-pick or close - it does not add any functionality and name probably can be better.


benjben:
Thanks @chuwy ! Cherry-picked in #481

PR #521, opened 2021-12-03
Enforce maximum field lengths (close #517)

### LINKED ISSUE #517: Validate enriched event against atomic schema before emitting
Events emitted by enrich should conform to [the atomic event schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0).  However, currently Enrich can emit events whose fields exceed the maximum allowed string length.  There should be an enrichment that converts events to failed events if they violate the protocol.

**Outstanding questions:**
Should the bad row be a Schema Violation or an Enrichment Failure?
Should it be configurably on/off or enforced on?

This change will remove the need for loaders to have special handling of long fields, [e.g. here](https://github.com/snowplow-incubator/snowplow-snowflake-loader/blob/0.8.2/transformer/src/main/scala/com/snowplowanalytics/snowflake/transformer/Transformer.scala#L92-L102)

voropaevp:
Enrichment uses the atomic event [schema](https://raw.githubusercontent.com/snowplow/iglu-central/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0) for shredded data to validate the pipeline output.
*This enrichment is enabled by default.* And has to be explicitly disabled to allow shredding of events that won't match with the atomic event schema.

Config has a single parameter `disable`. Which was left for backward compatibility with the old behaviour. 

Config schema: https://github.com/snowplow/iglu-central/pull/1194

Events that failed by this enrichment could return: 
- `EnrichmentFailure` with identifier `shredded_validator`, in case of error converting `EnrichedEvent` to the shredded format. 
- `SchemaViolation.IgluError` refering to `iglu:com.snowplowanalytics.snowplow/atomic/jsonschema/1.0.0`.

If no errors were raised - events are sent downstream, as they did before this change.

Benchmarks (10 sec run/ 10 iteration / 1 thread):

Validation *enabled*:

```
[info] # Warmup Iteration   1: 120.519 us/op
[info] Iteration   1: 108.637 us/op
[info] Iteration   2: 99.207 us/op
[info] Iteration   3: 99.795 us/op
[info] Iteration   4: 100.925 us/op
[info] Iteration   5: 103.841 us/op
[info] Iteration   6: 111.455 us/op
[info] Iteration   7: 289.276 us/op
[info] Iteration   8: 352.504 us/op
[info] Iteration   9: 125.896 us/op
[info] Iteration  10: 99.814 us/op
[info] Result "com.snowplowanalytics.snowplow.enrich.bench.EtlPipelineBench.measureProcessEventsIO":
[info]   149.135 ±(99.9%) 139.235 us/op [Average]
[info]   (min, avg, max) = (99.207, 149.135, 352.504), stdev = 92.095
```

Validation *disabled*:

```
[info] # Warmup Iteration   1: 125.246 us/op
[info] Iteration   1: 99.984 us/op
[info] Iteration   2: 100.672 us/op
[info] Iteration   3: 99.633 us/op
[info] Iteration   4: 99.737 us/op
[info] Iteration   5: 98.294 us/op
[info] Iteration   6: 97.941 us/op
[info] Iteration   7: 97.828 us/op
[info] Iteration   8: 97.971 us/op
[info] Iteration   9: 103.606 us/op
[info] Iteration  10: 152.935 us/op
[info] Result "com.snowplowanalytics.snowplow.enrich.bench.EtlPipelineBench.measureProcessEventsIO":
[info]   104.860 ±(99.9%) 25.675 us/op [Average]
[info]   (min, avg, max) = (97.828, 104.860, 152.935), stdev = 16.982
```

Evidently, the results are not stable and a little counterintuitive. Based on my measurement this is a performance increase, which can't be right. So there are probably issues either with test settings, test cleanups, or just a noise introduced by something completely unrelated like GC. But solving this mystery is outside of the scope for this issue.

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
@@ -16,26 +16,25 @@ package enrichments
 import java.nio.charset.Charset
 import java.net.URI
 import java.time.Instant
-
+import java.io.{PrintWriter, StringWriter}
 import org.joda.time.DateTime
-
 import io.circe.Json
-
 import cats.Monad
 import cats.data.{EitherT, NonEmptyList, OptionT, ValidatedNel}
 import cats.effect.Clock
 import cats.implicits._
 
+import com.snowplowanalytics.refererparser._
+
 import com.snowplowanalytics.iglu.client.Client
 import com.snowplowanalytics.iglu.client.resolver.registries.RegistryLookup
 
-import com.snowplowanalytics.iglu.core.SelfDescribingData
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer, SelfDescribingData}
 import com.snowplowanalytics.iglu.core.circe.implicits._
 
 import com.snowplowanalytics.snowplow.badrows._
 import com.snowplowanalytics.snowplow.badrows.{FailureDetails, Payload, Processor}
-
-import com.snowplowanalytics.refererparser._
+import com.snowplowanalytics.snowplow.badrows.FailureDetails.EnrichmentFailure
 
 import adapters.RawEvent
 import enrichments.{EventEnrichments => EE}
@@ -49,6 +48,8 @@ import outputs.EnrichedEvent
 import utils.{IgluUtils, ConversionUtils => CU}
 
 object EnrichmentManager {
+
+  val atomicSchema: SchemaKey = SchemaKey("com.snowplowanalytics.snowplow", "atomic", "jsonschema", SchemaVer.Full(1, 0, 0))
 
   /**
    * Run the enrichment workflow
@@ -92,6 +93,7 @@ object EnrichmentManager {
                enriched.pii = pii.asString
              }
            }
+      _ <- validateEnriched(enriched, raw, processor, client)
     } yield enriched
 
   /**
@@ -746,4 +748,55 @@ object EnrichmentManager {
       Failure.EnrichmentFailures(Instant.now(), fs),
       Payload.EnrichmentPayload(pee, re)
     )
+
+  private def validateEnriched[F[_]: Clock: Monad: RegistryLookup](
+    enriched: EnrichedEvent,
+    raw: RawEvent,
+    processor: Processor,
+    client: Client[F, Json]
+  ): EitherT[F, BadRow, Unit] =
+    enriched.toAtomicEvent
+      .leftMap(err =>
+        EnrichmentManager.buildEnrichmentFailuresBadRow(
+          NonEmptyList(
+            EnrichmentFailure(
+              None,
+              FailureDetails.EnrichmentFailureMessage.Simple(
+                "Error during conversion of enriched event to the atomic format"
+              )
+            ),
+            List(EnrichmentFailure(None, FailureDetails.EnrichmentFailureMessage.Simple(s"${cleanStackTrace(err)}")))
+          ),
+          EnrichedEvent.toPartiallyEnrichedEvent(enriched),
+          RawEvent.toRawEvent(raw),
+          processor
+        )
+      )
+      .toEitherT[F]
+      .flatMap(atomic =>
+        client
+          .check(SelfDescribingData(atomicSchema, atomic))
+          .leftMap(err =>
+            EnrichmentManager.buildEnrichmentFailuresBadRow(
+              NonEmptyList(
+                EnrichmentFailure(
+                  None,
+                  FailureDetails.EnrichmentFailureMessage.Simple(
+                    s"Enriched event not valid against ${atomicSchema.toSchemaUri}"
+                  )
+                ),
+                List(EnrichmentFailure(None, FailureDetails.EnrichmentFailureMessage.IgluError(atomicSchema, err)))
+              ),
+              EnrichedEvent.toPartiallyEnrichedEvent(enriched),
+              RawEvent.toRawEvent(raw),
+              processor
+            )
+          )
+      )
+
+  private def cleanStackTrace(t: Throwable): String = {
+    val sw = new StringWriter
+    t.printStackTrace(new PrintWriter(sw))
+    sw.toString
+  }
 }
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
@@ -15,8 +15,13 @@ package com.snowplowanalytics.snowplow.enrich.common.outputs
 import java.lang.{Integer => JInteger}
 import java.lang.{Float => JFloat}
 import java.lang.{Byte => JByte}
+import java.time.format.DateTimeFormatter
 
 import scala.beans.BeanProperty
+
+import cats.implicits._
+
+import io.circe.Json
 
 import com.snowplowanalytics.snowplow.badrows.Payload.PartiallyEnrichedEvent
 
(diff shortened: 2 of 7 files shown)


modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/ShreddedValidatorEnrichmentSpec.scala
```diff
@@ -0,0 +1,154 @@
...
+  )
+//  val oversized = CollectorPayload(api, Nil, None, None, source, context)
+
+  "config should parse" >> {
+    val configJson = json"""{
+      "disable": true
+      }"""
+    val schemaKey = SchemaKey(
+      "com.snowplowanalytics.snowplow.enrichments",
+      "shredded_validator_config",
```
istreeter, line 142: I'm struggling to think of a good name for this enrichment, and the best I can think of is `event_schema_validator_config` or `atomic_event_schema_validator_config` -- neither of which I like.

But I don't like `shredded_validator_config` either, and in particular the `shredded` bit.  I assume you called it this because of how the [atomic event iglu schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow/atomic/jsonschema/1-0-0) doesn't include event/context fields.  But my objections are:

* Enrich emits unshredded data.  In fact Enrich knows nothing about shredding data.
* This enrichment is helpful for people loading non-shredded data to a warehouse (as well as people loading shredded)
* Snowplow's strategy is to really push non-shredded formats going forwards.

benjben: I agree that `shredded` should not appear, as enrich should not worry about what comes next.
I would simply call it `enriched_validator_config`. I think that we don't need to refer to the schema in the name.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -16,28 +16,24 @@ package enrichments
...
 import com.snowplowanalytics.refererparser._
-
 import adapters.RawEvent
+import com.snowplowanalytics.snowplow.badrows.FailureDetails.{EnrichmentFailureMessage, EnrichmentInformation}
+import com.snowplowanalytics.snowplow.enrich.common.adapters.RawEvent.toRawEvent
+import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent.toPartiallyEnrichedEvent
+import com.snowplowanalytics.snowplow.enrich.common.utils.IgluUtils.buildSchemaViolationsBadRow
 import enrichments.{EventEnrichments => EE}
 import enrichments.{MiscEnrichments => ME}
 import enrichments.registry._
```
benjben, line 39: Not consistent in the packages

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -92,8 +88,53 @@ object EnrichmentManager {
...
+    enriched: EnrichedEvent,
+    raw: RawEvent,
+    processor: Processor,
+    client: Client[F, Json]
+  )(
+    implicit
+    M: Monad[F],
+    L: RegistryLookup[F],
+    C: Clock[F]
+  ): EitherT[F, BadRow, Unit] =
```
benjben, line 106:
```suggestion
  private def validateEnriched[F[_]: Clock: Monad: RegistryLookup](
    enriched: EnrichedEvent,
    raw: RawEvent,
    processor: Processor,
    client: Client[F, Json]
  ): EitherT[F, BadRow, Unit] =
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -92,8 +88,53 @@ object EnrichmentManager {
...
+      enriched.toShreddedEvent
+        .leftMap(err =>
+          EnrichmentManager.buildEnrichmentFailuresBadRow(
+            NonEmptyList.one(
+              FailureDetails.EnrichmentFailure(Some(EnrichmentInformation(atomicSchema, identifier = "shredded_validator")),
```
benjben, line 111:
```suggestion
              FailureDetails.EnrichmentFailure(Some(EnrichmentInformation(atomicSchema, identifier = "atomic_validator")),
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -92,8 +88,53 @@ object EnrichmentManager {
...
+                                               EnrichmentFailureMessage.Simple(err.getMessage)
+              )
+            ),
+            toPartiallyEnrichedEvent(enriched),
+            toRawEvent(raw),
+            processor
+          )
+        )
+        .toEitherT[F]
+        .flatMap(shreded =>
```
benjben, line 121:
```suggestion
        .flatMap(atomic =>
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
benjben, line 111: Would be nice to add a message in the bad row that says that the issue happened while validating the enriched event before sending.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -92,8 +88,53 @@ object EnrichmentManager {
...
+          client
+            .check(
+              SelfDescribingData(atomicSchema, shreded)
+            )
+            .leftMap(err =>
+              buildSchemaViolationsBadRow(
+                NonEmptyList.one(
+                  FailureDetails.SchemaViolation.IgluError(atomicSchema, err)
```
benjben, line 129: Would be nice to add a message in the bad row that says that the issue happened while validating the SDJ with atomic schema before sending the enriched event.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -92,8 +88,53 @@ object EnrichmentManager {
...
+          )
+        )
+        .toEitherT[F]
+        .flatMap(shreded =>
+          client
+            .check(
+              SelfDescribingData(atomicSchema, shreded)
+            )
+            .leftMap(err =>
+              buildSchemaViolationsBadRow(
```
benjben, line 127: In my opinion this should not be a schema violation. Schema violations are for the input SDJs that we receive from trackers. If we are here it means that the enrichments added/mutated something invalid, so this should be an enrichment failure.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
```diff
@@ -246,9 +247,163 @@ class EnrichedEvent extends Serializable {
...
+      Seq(
+        (k,
+         Json.fromString(
+           DateTimeFormatter.ISO_DATE_TIME.format(JsonSchemaDateTimeFormat.parse(v))
+         )
+        )
+      )
+    else Seq.empty[(String, Json)]
+
+  def toShreddedEvent: Either[Throwable, Json] =
```
benjben, line 269:
```suggestion
  def toAtomicEvent: Either[Throwable, Json] =
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
```diff
@@ -246,9 +247,163 @@ class EnrichedEvent extends Serializable {
...
   // Fields modified in PII enrichemnt (JSON String)
   @BeanProperty var pii: String = _
+
+  private val JsonSchemaDateTimeFormat =
+    DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss.SSS")
+
+  def toKv(k: String, v: String): Seq[(String, Json)] = if (v != null) Seq((k, Json.fromString(v))) else Seq.empty[(String, Json)]
+  def toKv(k: String, v: JByte): Seq[(String, Json)] = if (v != null) Seq((k, Json.fromBoolean(v == 1))) else Seq.empty[(String, Json)]
+  def toKv(k: String, v: JInteger): Seq[(String, Json)] = if (v != null) Seq((k, Json.fromInt(v))) else Seq.empty[(String, Json)]
+  def toKv(k: String, v: JFloat): Seq[(String, Json)] = if (v != null) Seq((k, Json.fromFloatOrNull(v))) else Seq.empty[(String, Json)]
```
benjben, line 257: You can write a generic version.

```suggestion
  def toKv[T](k: S, v: T, f: T => Json): Option[(String, Json)] =
    Option(v).map(value => (k, f(value)))
```
And then reuse it, e.g:
```
def toKv(k :String, b: JInteger): Option[(String, Json)]  = toKv(k, b, Json.fromInt)
```
You can call directly `.toSeq` on an `Option` so you can have e.g:
```
  toKv("user_id", this.user_id).toSeq
```
or put the `toSeq` directly in the above function

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/EnrichmentManagerSpec.scala
```diff
@@ -686,6 +686,28 @@ class EnrichmentManagerSpec extends Specification with EitherMatchers {
...
+      EnrichmentManager
+        .enrichEvent(
+          enrichmentReg,
+          client,
+          processor,
+          timestamp,
+          RawEvent(api, fatBody, None, source, context)
+        )
+        .map(_ => true)
+        .getOrElse(false) must beFalse
```
benjben, line 701: Would be great to do deeper testing with the content of the bad row

benjben (changes requested):
Enrich does not need to be aware of shredder/loader. What it is validating is the enriched event (atomic), not the shredded event, so `shredded" should not appear.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
voropaevp: I considered this, but there are issues with java/scala types. Working around them is more code, than current implementaion.
```
found   : Int => io.circe.Json
 required: Integer => io.circe.Json
  def toKv(k: String, v: JInteger): Seq[(String, Json)] = toKv[JInteger](k, v, Json.fromInt)

  found   : scala.Float => io.circe.Json
 required: java.lang.Float => io.circe.Json
  def toKv(k: String, v: JFloat): Seq[(String, Json)] = toKv[JFloat](k, v, Json.fromFloatOrNull)
  ```

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala (hunk 2 above)
Comment (voropaevp) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala:39]: Please elaborate, consistent with what?

Comment (benjben) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala]: In the packages that you added there is `com.snowplowanalytics.snowplow.enrich.common.` prefix whereas in the ones below there is not.

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala [hunk 11]
```diff
@@ -93,42 +92,51 @@ object EnrichmentManager {
...
-              FailureDetails.EnrichmentFailure(Some(EnrichmentInformation(atomicSchema, identifier = "atomic_validator")),
-                                               EnrichmentFailureMessage.Simple(err.getMessage)
+  ): EitherT[F, BadRow, Unit] = enriched.toAtomicEvent
+      .leftMap(err =>
+        EnrichmentManager.buildEnrichmentFailuresBadRow(
+          NonEmptyList(
+            EnrichmentFailure(
+              None,
+              FailureDetails.EnrichmentFailureMessage.Simple(
+                "Critical error during conversion of enriched event to the shredded format"
```
benjben, line 108: `shredded` => `atomic`

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -93,42 +92,51 @@ object EnrichmentManager {
...
+        client
+          .check(SelfDescribingData(atomicSchema, atomic))
+          .leftMap(err =>
+            EnrichmentManager.buildEnrichmentFailuresBadRow(
+              NonEmptyList(
+                EnrichmentFailure(
+                  None,
+                  FailureDetails.EnrichmentFailureMessage.Simple(
+                    s"Enriched event output validation error against the ${atomicSchema.toSchemaUri}. In most cases" +
+                      " it would be because of oversized fields."
```
benjben, line 129: I would not add `in most cases ...` because it can be misleading about the cause and anyway the error message from the validation should say what the problem is

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/EnrichmentManagerSpec.scala
```diff
@@ -697,8 +692,15 @@ class EnrichmentManagerSpec extends Specification with EitherMatchers {
...
-        .getOrElse(false) must beFalse
+        .swap
+        .forall {
+          case BadRow.EnrichmentFailures(_, failure, _) => failure.messages.map(_.message match {
+            case EnrichmentFailureMessage.Simple(error) => error.startsWith("Enriched event output validation error ")
+            case EnrichmentFailureMessage.IgluError(schemaKey, _) => schemaKey == SchemaKey("com.snowplowanalytics.snowplow", "atomic", "jsonschema", SchemaVer.Full(1, 0, 0))
+            case _ => false
+          }).foldLeft(true)(_ && _)
+          case _ => false
+        } must beTrue
```
benjben, line 703: :+1:

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/EnrichmentManagerSpec.scala
```diff
@@ -697,8 +692,15 @@ class EnrichmentManagerSpec extends Specification with EitherMatchers {
           timestamp,
           RawEvent(api, fatBody, None, source, context)
         )
-        .map(_ => true)
-        .getOrElse(false) must beFalse
+        .swap
+        .forall {
+          case BadRow.EnrichmentFailures(_, failure, _) => failure.messages.map(_.message match {
+            case EnrichmentFailureMessage.Simple(error) => error.startsWith("Enriched event output validation error ")
+            case EnrichmentFailureMessage.IgluError(schemaKey, _) => schemaKey == SchemaKey("com.snowplowanalytics.snowplow", "atomic", "jsonschema", SchemaVer.Full(1, 0, 0))
```
benjben, line 699: Neat: add error message

voropaevp: Specs matcher will print the error message. As of feedback round 3.

benjben: I meant to add a test on the error string contained in the `IgluError`

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/outputs/EnrichedEvent.scala
benjben: FYI I updated it

https://github.com/snowplow/enrich/pull/521/files#diff-a546d8f6e12feda9bb60528ceb0fc7160b3d388f429e0a2299e488fd044128aaR258-R274

PR #528, opened 2021-12-15
Release 2.0.4

dilyand:


project/Dependencies.scala
```diff
@@ -18,167 +18,167 @@ import sbt._
...
+    val postgresDriver = "42.2.16"
+    val mysqlConnector = "8.0.16"
+    val jaywayJsonpath = "2.4.0"
+    val iabClient = "0.2.0"
+    val yauaa = "5.23"
+    val guava = "28.1-jre"
+    val slf4j = "1.7.26"
+    val log4j = "2.16.0" // CVE-2021-44228
+
+    val refererParser = "1.1.0"
```
benjben, line 45: Can you remove the formatting please ?
It's going to be a massive pain to rebase `enrich-kinesis` otherwise.
Also this commit is about `Bump log4j-core to 2.16.0` and the whole file gets changed it's impossible to spot the change.

benjben (changes requested):
I think that you want to merge into `develop`

project/Dependencies.scala
benjben: If I `git clone` and run `sbt scalafmtAll` on the `master` branch format stays like this:
```
    val circe            = "0.13.0"
    val circeOptics      = "0.13.0"
    val circeConfig      = "0.7.0"
    val circeJackson     = "0.13.0"
    val scalaForex       = "1.0.0"
    val scalaWeather     = "1.0.0"
    val gatlingJsonpath  = "0.6.14"
    val scalaUri         = "1.4.5"
    val badRows          = "2.1.0"
    val igluClient       = "1.0.2"
```

dilyand: Hey @benjben , I got rid of the formatting changes but I had to completely turn off formatting using scalafmt for that. I think maybe some previous commits were done without formatting, which leads to conflicts.

When I check out `master` and run reformat file on Dependencies, it introduces many changes. But I find it a bit weird that `=` is not an alignment token in our `.scalafmt` file anyway...

Could you please green tick it when you're happy.

benjben: > When I check out master and run reformat file on Dependencies, it introduces many changes.
It might be that your local scalafmt config is used in the project, whereas only the one of the project should be used

benjben (approved):
:+1:

Issue #318, opened 2020-08-20
Add possibility to use STS to authenticate


benjben:
In order to be able to use [AWS Security Token Service](https://docs.aws.amazon.com/STS/latest/APIReference/welcome.html) to authenticate.

For example it can be used to pass AWS credentials with AWS_PROFILE environment variable.

Related to https://github.com/snowplow/snowplow/pull/4372

Lincon-Freitas:
Would be perfect to have the AssumeRoleWithWebIdentity available to allow me assuming roles directly in my pods.

PR #544, opened 2022-01-09
Enable AWS MSK IAM Authentication

CapChrisCap:
Hey, 

we are using the managed Kafka Service MSK from AWS. To enable the IAM authentication (see (https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html) instead of certificate based authentication, an additional package is required so that the authentication can be enabled in the configuration: 

```conf
[...]
# Or Kafka (Comment out for other types)
brokers = "...."
producerConf {
  "security.protocol"="SASL_SSL"
  "sasl.mechanism"="AWS_MSK_IAM"
  "sasl.jaas.config"="software.amazon.msk.auth.iam.IAMLoginModule required;"
  "sasl.client.callback.handler.class"="software.amazon.msk.auth.iam.IAMClientCallbackHandler"
}

consumerConf {
  "security.protocol"="SASL_SSL"
  "sasl.mechanism"="AWS_MSK_IAM"
  "sasl.jaas.config"="software.amazon.msk.auth.iam.IAMLoginModule required;"
  "sasl.client.callback.handler.class"="software.amazon.msk.auth.iam.IAMClientCallbackHandler"
}
```

This pull request contains the required package. 

We already built it with `sbt 'project streamKafka' docker:stage` and made it work in our test environments. 

Let me know if you have some questions or if there is something to improve :)


istreeter:
This looks good.  It's kinda similar to #318, which is also concerned with adding yet another different authentication method by adding a runtime dependency.  I don't see a problem with adding this one just for kafka.

I suggest adding the `% Runtime` modifier in the `project/Dependencies` file, a bit like [this example in the collector](https://github.com/snowplow/stream-collector/blob/c5d6452a9afd3b18630282b5f4285af4911e4474/project/Dependencies.scala#L60) because it's nice to be explicit about why we are adding the dependency if it's not needed to make the code compile.

benjben:
Thanks @CapChrisCap ! This will be released in next enrich version, we hope in the first half of February.

benjben:
Cherry-picked in https://github.com/snowplow/enrich/commit/143d19a51dc48261813bd0eedf1054536eba6c93

PR #545, opened 2022-01-11
enrich-kinesis: integration test

benjben:
The idea is to have an integration test (in `src/it/`) that
1) creates AWS resources (Kinesis streams)
2) starts enrich-kinesis
3) generates collector payloads on the fly, inserts them into Kinesis, reads the output of enrich, and counts the events
4) stops enrich-kinesis
5) destroys the resources

Resources creation/deletion is performed [here](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/test/Resources.scala#L33)

Generation of data and aggregation of the output are performed [here](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/test/Main.scala#L32-L53)

What is missing is to start and stop enrich. My idea was to call directly [run(args)](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala#L54) and specify args that point to [this config file](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/resources/config.hocon) and [this resolver](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/resources/iglu_resolver.json) (by retrieving their path dynamically with `getResource()`). This `IO` needs to be started in the background and to be stopped after we are done with sending data.

Previous [Main](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/test/Main.scala#L21-L25) needs to be updated to run the 5 steps. It doesn't need to receive args any more. [These values](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/test/Resources.scala#L26-L31) should be used everywhere in the test.

[integrationTests.sh](https://github.com/snowplow/enrich/blob/feature/integration-test/modules/kinesis/src/it/resources/integrationTests.sh) is deprecated


benjben:
Provided that AWS environment variables are set, the test can be run with `sbt "project kinesis" IntegrationTest/test`

PR #552, opened 2022-02-15
retry using jitter & don't crash in case of KPL error

oguzhanunlu:


modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -122,7 +122,7 @@ object Sink {
     producer: KinesisProducerClient[F],
     data: AttributedData[Array[Byte]]
   ): F[Unit] = {
-    val retryPolicy = capDelay[F](config.backoffPolicy.maxBackoff, exponentialBackoff[F](config.backoffPolicy.minBackoff))
+    val retryPolicy = capDelay[F](config.backoffPolicy.maxBackoff, fullJitter[F](config.backoffPolicy.minBackoff))
```
benjben, line 125: Will this be retried forever?

oguzhanunlu: yeah, `exponentialBackoff` retries forever as well, the only difference is jitter

benjben: Right. So only removing `Async[F].raiseError(exception)` below makes sure that the retries are done forever ?

oguzhanunlu: I think so but I'm doing another stress test now, I'll update here once I'm done with it, so that we are confident in the change.

oguzhanunlu: I can confirm that removing raiseError here allows retries to continue forever

benjben: Awesome !

oguzhanunlu: great finding, then it is a decision of who should control retry policy, our codebase via cats-retry (when recordTtl is low) or KPL (when recordTtl is high)? in both cases we'll have indefinite retry.

benjben: As I see it it's better to use directly KPL for the retry, as it removes a layer

benjben:
Hey @oguzhanunlu , I [added](https://github.com/snowplow/enrich/pull/552/commits/dbe8214189f6b6ca7ef10caedec9d0a95cebf0a1) a few changes:
- [KPL retries forever](https://github.com/snowplow/enrich/pull/552/commits/dbe8214189f6b6ca7ef10caedec9d0a95cebf0a1#diff-8f23b8409885a2558d6def2d0746935cd382d72eedad985dc426dfd5d71b0027R100) (well, for a very very long time)
- [Better logging](https://github.com/snowplow/enrich/pull/552/commits/dbe8214189f6b6ca7ef10caedec9d0a95cebf0a1#diff-8f23b8409885a2558d6def2d0746935cd382d72eedad985dc426dfd5d71b0027R150) in case of KPL failure
- [Better logging](https://github.com/snowplow/enrich/pull/552/commits/dbe8214189f6b6ca7ef10caedec9d0a95cebf0a1#diff-8f23b8409885a2558d6def2d0746935cd382d72eedad985dc426dfd5d71b0027R155) in case of KPL error
- [streamExists](https://github.com/snowplow/enrich/pull/552/commits/dbe8214189f6b6ca7ef10caedec9d0a95cebf0a1#diff-61bc9cf07ea8e59b2053b6f69672e14f366671795a6e5414761621f7635168e5R29) function that is run both for sink and source

If that's all fine with you I'll merge the PR

oguzhanunlu (commented):
GitHub naturally doesn't allow me to approve my own PR but it looks good to me!

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -139,10 +145,15 @@ object Sink {
         wasSuccessful = _.isSuccessful,
         policy = retryPolicy,
         onFailure = (result, retryDetails) =>
-          Logger[F].warn(s"Writing to shard ${result.getShardId()} failed after ${retryDetails.retriesSoFar} retry"),
+          Logger[F]
+            .warn(
+              s"Writing to shard ${result.getShardId()} failed after ${retryDetails.retriesSoFar} retries. Errors : [${getErrorMessages(result)}]"
```
oguzhanunlu, line 150: just a nit, since we let KPL retry forever, `retryDetails.retriesSoFar` will always be 0 until KPL retries Long.MaxValue times

benjben: I should make it clear what is KCL retry and what is cats-retry retry

benjben: > since we let KPL retry forever, retryDetails.retriesSoFar will be 0 until KPL retries
I wonder if we can be 100% sure that KPL won't fail for some unexpected reason, so I would still let the number of cats-retry retries, what do you think ?

Please see https://github.com/snowplow/enrich/pull/552/commits/6794e4fb8c84a79eb9db7ff47bf8adaf3daaa1b4

oguzhanunlu: >  KPL won't fail for some unexpected reason

I'd prefer to protect ourselves against that so I'd let cats-retry retry forever as well and report both retry numbers in log message

benjben:
cherry-picked in #481

PR #570, opened 2022-03-15
common-fs2: add support for remote adapters

oguzhanunlu:
- first commit introduces support for remote adapter
- second commit introduces blocker not to block compute pool

### DIFF
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
@@ -53,9 +53,6 @@ import com.snowplowanalytics.snowplow.enrich.common.utils.ConversionUtils
 
 object Enrich {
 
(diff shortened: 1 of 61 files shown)


istreeter (commented):
Well done @oguzhanunlu!

This implementation will work well.  It does the most important thing, which is to avoid blocking the main threads.  It will let you do a fair test of this implementation vs using a http4s client.

I think there is a way you could have made it simpler.  If you added `BlockerF` as a field of [the RemoteAdapter case class](https://github.com/snowplow/enrich/blob/3.0.0/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/RemoteAdapter.scala#L41-L43) then you would not need to change the signature of `toRawEvents`.  It would mean changing fewer files.

_An aside:_

I look forward to when we deprecate stream-enrich and then use cats-effect 3 throughout the core module.  That will mean we don't need to pass a `BlockerF` in any signature.

Before we do this, we need to write fs2 sources/sinks for nsq and kafka.

oguzhanunlu:
scalaj got outperformed by http4s in #571 , hence this is not to be merged

PR #571, opened 2022-03-17
common-fs2: support remote adapter via http4s

oguzhanunlu:


modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/AssetsSpec.scala
```diff
@@ -116,38 +119,41 @@ class AssetsSpec extends Specification with CatsIO with ScalaCheck {
...
-
-          val after = Stream.eval(state.hashes.get.map(_.get(uri)))
-          before ++ update ++ after
-        }
-        .withHttp
-        .haltAfter(3.second)
-        .compile
-        .toList
-        .map(_ == List(Some(Assets.Hash(maxmind1Hash)), (), Some(Assets.Hash(maxmind2Hash))))
+      Clients.mkHttp[IO](ExecutionContext.global).use { httpClient =>
```
istreeter, line 122: You might find it interesting to look at [these unit tests in iglu-scala-client](https://github.com/snowplow/iglu-scala-client/blob/e9b8b9c2718a45da0ce4c25d50cea54bdac5dfcf/modules/http4s/src/test/scala/com.snowplowanalytics.iglu/client/resolver/registries/Http4sRegistryLookupSpec.scala#L39-L45).  Remember a `Client[F]` is just a kleisli around a `Request[F] => F[Response[F]]` so you can easily implement something custom in a unit test.  You don't need to instantiate a full Blaze client with an underlying tcp pool.

Here you instantiate a full blaze client

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/HttpClient.scala
```diff
@@ -13,28 +13,80 @@
...
+        uri: String,
+        authUser: Option[String],
+        authPassword: Option[String],
+        body: Option[String],
+        method: String,
+        connectionTimeout: Option[Long],
+        readTimeout: Option[Long]
+      ): F[Either[Throwable, String]] = {
+        val request = Request[F](
+          uri = Uri.unsafeFromString(uri),
```
istreeter, line 51: I know this _should_ be safe, but I don't like seeing things like `unsafeFromString` in the code!  Can you think of any ways to avoid it?

oguzhanunlu: good point, I replaced that with a safer option

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/AssetsSpec.scala
istreeter: That wasn't what I meant 😆.  My point was that in a _unit test_ your client does not need to make real http requests whatsoever.  You can "fake" a client by making it respond with whatever response you want for the test.

```
val client = HttpClient[IO] { _ =>
  Resource.eval(Ok("ok!"))
}
```

It's not a huge problem -- I am only making this suggestion to make you aware of the testing pattern, because I think it is a nice one.

istreeter (commented):
Please can you explain the reasons for making the adapters into classes into objects?  I can see you now pass the capability traits (e.g. `RegistryLookup` and `Clock`) during construction of the adapter, instead of when you call the `toRawEvents` function.

This is a very opinionated change!!

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/AssetsSpec.scala
oguzhanunlu: ah, my bad, I see the idea of mocking, that's nice, I'll use it :)

istreeter: I don't think of it as mocking.  I think of it as honouring the trait.  In http4s, a client **is** a function from `Request[F]` to `Response[F]`.  And if the function under test takes a client in its signature, then we should be allowed to give it any chosen implementation of `Request[F] => F[Response[F]]`.

To me, mocking means something slightly different.  I think mocking is when you hack at a class to make it pretend to be something that it is not.

oguzhanunlu: Replaced all similar uses with thin http clients similar to iglu scala client

oguzhanunlu:
hey @istreeter @benjben , after recent changes, PR reduced from touching 84 files to 29 files.

Good to know before review:
- I intentionally didn't touch Stream Enrich at all
- common-fs2 has a different remote adapter configuration structure compared to Stream Enrich
- Apart from remote adapter changes, this PR also removes usage of blocker from api request enrichment since http4s itself is non-blocking. Removing blocker from signatures increased # of files here.
- Tests use a thin http4s client based on iglu-scala-client tests

I'll think about logging in HttpClient, especially not logging with every failure, however I wanted to start code review phase

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala
```diff
@@ -54,6 +55,7 @@ final case class ConfigFile(
   output: Outputs,
   concurrency: Concurrency,
   assetsUpdatePeriod: Option[FiniteDuration],
+  remoteAdapters: Option[RemoteAdapterConfigs],
```
istreeter, line 58: How about make this non-optional?  The default value in `application.conf` could be:
```
"remoteAdapters": {
  "connectionTimeout": "10 seconds"
  "readTimeout": "45 seconds"
  "maxConnections": 10
  "configs: []
}
```

From point of view of the user it is still optional.  Meaning if the user does not care about remote adapters then they are allowed to not put anything extra in their config hocon.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -252,4 +262,25 @@ object Environment {
...
+   * @param remoteAdaptersConfig List of configuration per remote adapter
+   * @return An Http4sClient with a mapping of vendor-version and the adapter assigned for it
+   */
+  def prepareRemoteAdapters[F[_]: ConcurrentEffect](
+    remoteAdaptersConfig: Option[RemoteAdapterConfigs],
+    ec: ExecutionContext
+  ): Resource[F, (Http4sClient[F], Map[(String, String), RemoteAdapter])] =
+    remoteAdaptersConfig.fold(
+      Clients.mkHttp(None, None, None, ec).map((_, Map.empty[(String, String), RemoteAdapter]))
+    ) { adapters =>
```
istreeter, line 277: This is an example of some code that is hard to understand, but would be much easier to understand if `RemoteAdapterConfigs` is non-optional.

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/test/TestEnvironment.scala
```diff
@@ -111,16 +114,23 @@ object TestEnvironment extends CatsIO {
...
+  val adapterRegistry = new AdapterRegistry()
+
   val igluClient: Client[IO, Json] =
     Client[IO, Json](Resolver(List(embeddedRegistry), None), CirceValidator)
 
+  val http4sClient: Http4sClient[IO] = Http4sClient[IO] { _ =>
+    val dsl = new Http4sDsl[IO] {}; import dsl._
+    Resource.eval(Ok(""))
+  }
+
```
istreeter, line 126: 👍

istreeter (commented):
This is an excellent PR @oguzhanunlu !  There is some really good stuff in here.  It's very neat, and it uses the http4s client very well.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Assets.scala
```diff
@@ -154,7 +154,7 @@ object Assets {
    * Creates an update stream that periodically checks if new versions of assets are available.
    * If that's the case, updates them locally for the enrichments and updates the state.
    */
-  def updateStream[F[_]: ConcurrentEffect: ContextShift: Parallel: Timer](
+  def updateStream[F[_]: ConcurrentEffect: ContextShift: Parallel: Timer: HttpClient](
```
benjben, line 157: As I understand, `HttpClient` is an abstraction of the underlying HTTP client. But in many places we specify directly `Http4sClient` instead of `HttpClient`. I assumed that we do so for code that is used only in enrich fs2 and not in stream enrich. But here this code is used only for enrich fs2 and we use `HttpClient` instead of `Http4sClient`. Is there a reason?

config/config.pubsub.extended.hocon
```diff
@@ -111,6 +111,16 @@
   # no assets will be updated if the key is absent
   "assetsUpdatePeriod": "7 days"
 
+  # Optional, configuration of remote adapters
+  # Increasing maxConnections is helpful if your adapter has high latency
+  # remoteAdapters.configs is a list of adapter configurations
+  "remoteAdapters": {
+    "connectionTimeout": "10 seconds",
+    "readTimeout": "45 seconds",
+    "maxConnections": 10,
```
benjben, line 120: Neat: put these in `application.conf` so that they are all optional and it's possible to specify only `configs`

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -93,9 +94,11 @@ final case class Environment[F[_], A](
   enrichments: Ref[F, Environment.Enrichments[F]],
   semaphore: Semaphore[F],
   assetsState: Assets.State[F],
-  httpClient: HttpClient[F],
+  httpClient: Http4sClient[F],
+  remoteAdapterHttpClient: Http4sClient[F],
```
benjben, line 98:
```suggestion
```
See the other comment, the `Environment` doesn't need to hold this Http4s client, because `HttpClient` could be hold directly inside a `RemoteAdapter`, and also the `Resource` could be created directly inside `Environment.make` so that the HTTP client is automatically released when `Environment` terminates.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -56,8 +58,17 @@ object Clients {
       }
     }
 
-  def mkHttp[F[_]: ConcurrentEffect](ec: ExecutionContext): Resource[F, HttpClient[F]] =
-    BlazeClientBuilder[F](ec).resource
+  def mkHttp[F[_]: ConcurrentEffect](
+    connectionTimeout: FiniteDuration = defaults.ConnectTimeout,
+    readTimeout: FiniteDuration = defaults.RequestTimeout,
+    maxConnections: Int = 10, // http4s uses 10 by default
```
benjben, line 64: It would be awesome to make `connectionTimeout`, `readTimeout` and `maxConnections` configurable with defaults and to explain in the config file what the HTTP client is used for.

I know that we discussed using the lib defaults vs using our own. Here already we are using one of our own (`10`) that is impossible to update. What if we want to update the timeouts also? I'm strongly in favor of putting these in `application.conf`. Default timeouts from the lib will probably never change and even if they do and we didn't update them in `application.conf`, well, they are just defaults and can still be updated in the `config.hocon`.

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
```diff
@@ -256,6 +274,75 @@ class ConfigFileSpec extends Specification with CatsIO {
       }
     }
 
+    "parse valid adapter registries" in {
```
benjben, line 277: You could put one remote adapter config in the sample in `config/`. This way you document it for OSS users, and also you can get rid of this new test, that would require to be updated whenever we update the config.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -169,24 +172,31 @@ object Environment {
...
       assets = parsedConfigs.enrichmentConfigs.flatMap(_.filesToCache)
+      (remoteAdapterHttpClient, remoteAdapters) <- prepareRemoteAdapters[F](file.remoteAdapters, ec)
+      adapterRegistry = new AdapterRegistry(remoteAdapters)
       sem <- Resource.eval(Semaphore(1L))
       assetsState <- Resource.eval(Assets.State.make[F](blocker, sem, clts, assets))
-      enrichments <- Enrichments.make[F](parsedConfigs.enrichmentConfigs, BlockerF.ofBlocker(blocker))
+      enrichments <- {
+        implicit val C: Http4sClient[F] = http
+        Enrichments.make[F](parsedConfigs.enrichmentConfigs, BlockerF.ofBlocker(blocker))
+      }
```
benjben, line 188: Is there a reason why you use `HttpClient` implicitly everywhere instead of just passing it as a regular parameter? It could ease the reading of the code

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala
```diff
@@ -147,8 +149,14 @@ object Run {
...
-      val enrich = Enrich.run[F, A](env)
-      val updates = Assets.run[F, A](env.blocker, env.semaphore, env.assetsUpdatePeriod, env.assetsState, env.enrichments)
+      val enrich = {
+        implicit val remoteAdapterHttpClient: Http4sClient[F] = env.remoteAdapterHttpClient
+        Enrich.run[F, A](env)
+      }
+      val updates = {
+        implicit val httpClient: Http4sClient[F] = env.httpClient
+        Assets.run[F, A](env.blocker, env.semaphore, env.assetsUpdatePeriod, env.assetsState, env.enrichments)
+      }
```
benjben, line 159:
```suggestion
implicit val httpClient: Http4sClient[F] = env.httpClient
val enrich = Enrich.run[F, A](env)
val updates = Assets.run[F, A](env.blocker, env.semaphore, env.assetsUpdatePeriod, env.assetsState
```
Don't only the remote adapters need `remoteAdaptersHttpClient` ?

benjben (commented):
Sorry for all the comments, it looks great overall :+1:

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -252,4 +262,24 @@ object Environment {
...
+   * @param remoteAdaptersConfig List of configuration per remote adapter
+   * @return An Http4sClient with a mapping of vendor-version and the adapter assigned for it
+   */
+  def prepareRemoteAdapters[F[_]: ConcurrentEffect](
+    remoteAdapters: RemoteAdapterConfigs,
+    ec: ExecutionContext
+  ): Resource[F, (Option[Http4sClient[F]], Map[(String, String), RemoteAdapter])] = {
+    val preparedRemoteAdapters =
+      remoteAdapters.configs.map { config =>
+        (config.vendor, config.version) -> RemoteAdapter(config.url, None, None)
```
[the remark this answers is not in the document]
oguzhanunlu: they are [used](https://github.com/snowplow/enrich/blob/master/modules/stream/common/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/Enrich.scala#L327-L328) in Stream Enrich

benjben (approved):

oguzhanunlu:
metrics commit is approved at https://github.com/snowplow/enrich/pull/600 by Ian

istreeter (approved):
Approved!  This will work.

However, I wonder if the code would have looked overall neater if you set things up slightly differently:

First, change [these counters](https://github.com/snowplow/enrich/blob/aafa991814bf21a1bcf71ed21791e02eeb6877fa/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Metrics.scala#L79-L81) so they are `Int` not `Option[Int]`.  Then instead of passing `remoteAdaptersEnabled` into the `MetricsRefs.init` method, you instead pass it down to `StatsDReporter.make` and `stdoutReporter`.  Filter out metrics in the reporters not the counters.  You _might_ find this removes a lot of the Option flatMapping.

You're welcome to play around with that idea if you want, but I'm also happy to let you merge it with the current design :+1:

oguzhanunlu:
all commits except the last one https://github.com/snowplow/enrich/pull/571/commits/499f98f383cbd52c94077c0b85e0afacc9e90702 is reviewed and approved, I force-pushed to resolve conflicts with develop

oguzhanunlu:
hey @spenes , feel free to review the whole PR but I'd like to get your review especially for the latest commit

Downstream in the BRA, we want to aggregate adapter failures, not all of them but the ones caused by remote adapter failures. After talking to @istreeter , I decided to be pragmatic over perfectionist here (ideally a new version of adapter_failure bad row supporting remote adapter failures and downstream updates starting with badrows lib...) and used a straightforward prefix to distinguish adapter failures we want to aggregate.

I tested the functionality on sandbox succcessfully

oguzhanunlu:
rebasing on latest develop broke the PR due to switching from http4s client to ember, I'll fix it

spenes (approved):
Last commit looks good to me 👍

oguzhanunlu:
reverted https://github.com/snowplow/enrich/commit/c010f1266c4d35c8324e739bfd2c984344ae35cc after seeing Ember not supporting timeout options of Blaze using `git revert`, followed by a squash of previously reviewed & approved commits

PR #594, opened 2022-04-06
Set fixed termination period for pre-shutdown hook

peel:
Previously it was possible for pre-shutdown hook to take longer than expected resulting in a slow shutdown.
This affected integration test for kinesis and may have extended graceful shutdown time if the metadata endpoint was very slow to respond.

### DIFF
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
@@ -90,7 +90,7 @@ object Enrich {
       _.parEvalMap(env.streamsSettings.concurrency.sink)(sinkChunk(_, sinkOne(env), env.metrics.enrichLatency))
         .evalMap(env.checkpoint)
 
-    Stream.eval(runWithShutdown(enriched, sinkAndCheckpoint, env.preShutdown.getOrElse(() => Sync[F].unit)))
+    Stream.eval(runWithShutdown(enriched, sinkAndCheckpoint)).onComplete(env.preShutdown)
   }
 
   /**
@@ -262,8 +262,7 @@ object Enrich {
    */
   private def runWithShutdown[F[_]: Concurrent: Sync: Timer, A](
     enriched: Stream[F, List[(A, Result)]],
-    sinkAndCheckpoint: Pipe[F, List[(A, Result)], Unit],
-    preShutdown: () => F[Unit]
+    sinkAndCheckpoint: Pipe[F, List[(A, Result)], Unit]
   ): F[Unit] =
     Queue.synchronousNoneTerminated[F, List[(A, Result)]].flatMap { queue =>
       queue.dequeue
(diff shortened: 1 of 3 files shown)


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -278,15 +278,15 @@ object Enrich {
...
             // Runtime exception in the stream of enriched events.
             // We wait for the enriched events already in the queue to get sunk and checkpointed.
             // We then raise the original exception
             Logger[F].error(e)("Unexpected error in enrich") *>
-              preShutdown() *> terminateStream(queue, fiber).handleErrorWith { e2 =>
-              Logger[F].error(e2)("Error when terminating the stream")
-            } *> Sync[F].raiseError(e)
+              terminateStream(queue, fiber).handleErrorWith { e2 =>
+                Logger[F].error(e2)("Error when terminating the stream")
+              } *> preShutdown() *> Sync[F].raiseError(e)
```
benjben, line 289: Why has the order some importance ?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -204,7 +204,7 @@ object Environment {
       getRegionFromConfig(file).orElse(getRegion),
       cloud,
       acceptInvalid,
-      Some(() => metadata.submit.compile.drain)
+      Some(() => metadata.submit.interruptAfter(1.minute).compile.drain)
```
benjben, line 207: 1. Why is `preShutdown` an `Option` ? Is it not always `Some` ?
2. In the integration test, I doubt that the issue was that `preShutdown` was taking longer than 1 minute, it would rather seem that it would get stuck forever
3. If I understand correctly, the reason why we need this `preShutdown` is that unlike all the other streams [here](https://github.com/snowplow/enrich/blob/master/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala#L151-L153), we want to flush the buffer of metadata before exiting. Ideally, we would want that to be automatically triggered when [enrich](https://github.com/snowplow/enrich/blob/master/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala#L150) stream terminates. What about using `enrich.onComplete()` to flush the metadata ? This way you wouldn't need to have this `preShutdown` in the `Environment` and to worry about it. The flushing needs to be a function and not a stream.

peel: That's a really good architectural point!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -204,7 +204,7 @@ object Environment {
       getRegionFromConfig(file).orElse(getRegion),
       cloud,
       acceptInvalid,
-      Some(() => metadata.submit.compile.drain)
+      metadata.submit.interruptAfter(1.minute)
```
benjben, line 207: I'm still not at ease with the fact that the shutdown will wait for an infinite stream to be interrupted 1 minute later (unless I'm missing something?). What about having `def flush: F[Unit]` [in here](https://github.com/snowplow/enrich/blob/3.1.2/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/experimental/Metadata.scala#L46-L50) that sends what's in `MetadataEventsRef.snapshot` when called ?

peel: Why infinite? We take a snapshot, then for all events we execute HTTP requests. This means a finite stream. The interruption will stop if metadata submitting takes longer than that which means that upstream endpoint is not responding and metadata cannot be submitted. I think capping that makes sense.

istreeter: I managed to get the integration test to pass by changing [this line](https://github.com/snowplow/enrich/blob/master/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/experimental/Metadata.scala#L87) to `Stream.empty[F]` instead of `Stream.never[F]`.  No other change needed.  I think semantically this might be the correct fix: if it's a `noop` implementation of metadata then `submit` should return immediately, not never.

PR #600, opened 2022-04-12
Integrate metrics to remote adapters

istreeter:


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -279,7 +282,24 @@ object Environment {
...
+      for {
+        now <- Resource.eval(Clock[F].realTime(MILLISECONDS))
+        resp <- inner.run(req)
+        _ <- Resource.eval {
+               metrics.remoteAdaptersLatency(true, now) *>
+                 (resp.status.responseClass match {
+                   case Status.Successful => metrics.remoteAdapters2xxCount(true)
+                   case Status.ServerError => metrics.remoteAdapters5xxCount(true)
+                   case _ => Sync[F].unit
+                 })
```
istreeter, line 301: It would be great to capture timeouts, or failed to connect.  When you call `inner.run(req)` you only successfully get a response if the http call ends with a response.  But there is a whole different class of http error, where you don't get back a response at all!

I think you need something like `.handleErrorWith()` in order to capture the timeout exception.

istreeter (commented):
Looks good @oguzhanunlu, I like how you capture latency.

Just two suggestions / ideas.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Metrics.scala
```diff
@@ -132,14 +147,34 @@ object Metrics {
       def invalidCount: F[Unit] =
         refsStatsd.invalidCount.update(_ + 1) *>
           refsStdout.invalidCount.update(_ + 1)
+
+      def remoteAdapters2xxCount(enabled: Boolean): F[Unit] =
+        if (enabled) refsStatsd.remoteAdapters2xxCount.update(_.map(_ + 1)) *> refsStdout.remoteAdapters2xxCount.update(_.map(_ + 1))
```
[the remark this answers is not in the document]
oguzhanunlu: right! it is not needed, removed in a separate commit

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -279,7 +283,27 @@ object Environment {
...
     else Resource.pure[F, (Option[Http4sClient[F]], Map[(String, String), RemoteAdapter])]((None, preparedRemoteAdapters))
   }
+
+  def enableRemoteAdapterMetrics[F[_]: Clock: Sync](metrics: Metrics[F])(inner: Http4sClient[F]): Http4sClient[F] =
+    Http4sClient { req =>
+      for {
+        now <- Resource.eval(Clock[F].realTime(MILLISECONDS))
+        resp <- inner.run(req).recoverWith {
+                  case _: TimeoutException =>
+                    Resource.eval(metrics.remoteAdaptersTimeoutCount.map(_ => Response.timeout[F]))
```
istreeter, line 297: I think this will increment the timeout count **and** the 5xx count.

Because you catch the exception and replace it with `Response.timeout[F]` which [is a regular response with a 503 status code](https://github.com/http4s/http4s/blob/v0.21.33/core/src/main/scala/org/http4s/Message.scala#L606).  So then `resp.status.responseClass` will match `Status.ServerError` on line 303.

I don't know if this was your intention, but my biggest complaint is **it is not obvious just by looking at the code**.

There are a few possible alternatives, but my favourite is to increment the timeout count and then let the exception bubble up:

```
resp <- inner.run(req).attemptTap {
      case Left(_: TimeoutException) => Resource.eval(metrics.remoteAdaptersTimeoutCount)
      case _ => Resource.pure[F, Unit](())
    }
```


Furthermore, you catch only timeout exceptions, but other exceptions are possible.  For example, if the api returns invalid bytes, or if the TLS handshake fails, then you get a different exception.  Maybe it's worth talking to techops about which metric should be incremented in these cases.  Currently you are not incrementing any metric if this happens.

oguzhanunlu: It was my intention but I wasn't semantically sure about it. I'll talk to TechOps about that and update this accordingly. Thanks for the feedback!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -279,7 +282,24 @@ object Environment {
...
+    Http4sClient { req =>
+      for {
+        resp <- inner.run(req).attemptTap {
+                  case Left(_: TimeoutException) => Resource.eval(metrics.remoteAdaptersTimeoutCount)
+                  case Left(_) => Resource.eval(metrics.remoteAdaptersFailureCount)
+                  case Right(response) =>
+                    response.status.responseClass match {
+                      case Status.Successful => Resource.eval(metrics.remoteAdaptersSuccessCount)
+                      case Status.ServerError => Resource.eval(metrics.remoteAdaptersFailureCount)
+                      case _ => Resource.pure[F, Unit](())
```
istreeter, line 300: What type of response does this last case match?  Currently this response is not captured by metrics, so there is no visibility to the pipeline operator.

oguzhanunlu: Last one corresponds to all responses except 2xx & 5xx (1xx, 3xx and 4xx). I think I should include 4xx in `remoteAdaptersFailureCount` as well, as it is a failure just like 5xx. However 1xx and 3xx aren't failures semantically.

oguzhanunlu: There is an inconsistency between what I do here and what's done at HttpClient Sync instance in terms of status code interpretation, I'll align both places

oguzhanunlu: I just sent a new commit to align both places, now any response without 2xx status code will be adapter failure and reported in `remoteAdaptersFailureCount` and any response with 2xx status code will be reported in `remoteAdaptersSuccessCount` where all cases are covered and made visible to the pipeline operator

wdyt @istreeter ?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -279,7 +282,21 @@ object Environment {
...
   }
+
+  def enableRemoteAdapterMetrics[F[_]: Sync](metrics: Metrics[F])(inner: Http4sClient[F]): Http4sClient[F] =
+    Http4sClient { req =>
+      for {
+        resp <- inner.run(req).attemptTap {
+                  case Left(_: TimeoutException) => Resource.eval(metrics.remoteAdaptersTimeoutCount)
+                  case Left(_) => Resource.eval(metrics.remoteAdaptersFailureCount)
+                  case Right(response) =>
+                    if (response.status.responseClass == Status.Successful) Resource.eval(metrics.remoteAdaptersSuccessCount)
```
istreeter, line 297: The problem with the `==` operator is it's not very type-safe.  Whenever I use `==` I worry that I will make a coding mistake and the compiler won't help me out by telling me.  You might find it interesting to read [this description of the problem from Cats](https://typelevel.org/cats/typeclasses/eq.html).

In this case I would probably have used a `match`, which I know is type safe so I trust the compiler will tell me if I get it wrong

```
    response.status.responseClass match {
      case Status.Successful => Resource.eval(metrics.remoteAdaptersSuccessCount)
      case _ => Resource.eval(metrics.remoteAdaptersFailureCount)
    }
```

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -279,7 +282,21 @@ object Environment {
...
       Clients
         .mkHttp(remoteAdapters.connectionTimeout, remoteAdapters.readTimeout, remoteAdapters.maxConnections, ec)
+        .map(enableRemoteAdapterMetrics(metrics))
         .map(c => (c.some, preparedRemoteAdapters))
     else Resource.pure[F, (Option[Http4sClient[F]], Map[(String, String), RemoteAdapter])]((None, preparedRemoteAdapters))
   }
+
+  def enableRemoteAdapterMetrics[F[_]: Sync](metrics: Metrics[F])(inner: Http4sClient[F]): Http4sClient[F] =
+    Http4sClient { req =>
+      for {
```
istreeter, line 292: Just a bit of tidy-up, but you can now remove the for-loop.  Wherever you have

```
for {
  a <- something
} yield a
```

you can replace it with just

```
something
```

istreeter (commented):
Excellent!  I completely agree with your decision to align the metrics with how the http client uses the response.  Now the metrics tell the pipeline operator whether the remote adapter was successful, which is the most important thing to know.

I'm approving this because it's good enough already!  I made a couple of comments, but only because I care about your scala education 😉

oguzhanunlu:
thanks @istreeter ! let me address them as well :)

oguzhanunlu:
merged to `http4s-remote-adapter` at https://github.com/snowplow/enrich/pull/571

PR #618, opened 2022-06-05
Common: Return to original intended enrichment order

istreeter:


benjben (approved):
This is brilliant !

I just wonder about the feature flag, I don't see a reason why we would not want to preserve the original order (until we have the order of enrichments configurable, which will be way easier to implement with this PR).

Issue #626, opened 2022-06-13
Common: fix mysql connection


oguzhanunlu:
Tested sql query enrichment using enrich-kinesis 3.1.5 and whether ssl is enabled or not in the enrichment config, it fails as following:

```
[ioapp-compute-0] INFO com.zaxxer.hikari.HikariDataSource - HikariPool-1 - Starting...
[ioapp-compute-0] ERROR com.zaxxer.hikari.pool.HikariPool - HikariPool-1 - Exception during pool initialization.
com.mysql.cj.jdbc.exceptions.CommunicationsException: Communications link failure
The last packet sent successfully to the server was 0 milliseconds ago. The driver has not received any packets from the server.
...
Caused by: com.mysql.cj.exceptions.CJCommunicationsException: Communications link failure
...
Caused by: javax.net.ssl.SSLHandshakeException: No appropriate protocol (protocol is disabled or cipher suites are inappropriate)
```

https://github.com/snowplow/enrich/blob/c50439d2da24e62f47345cec0f497e93cb0729d8/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala#L64-L68

Enrich doesn't set `useSsl` when enrichment config has `sslMode=false` and by default useSsl is true in connector library. It means our config doesn't get reflected in connection as expected and it always uses encrypted connections.

oguzhanunlu:
We can fix this by explicit TLS version flag but connector lib also has a newer version which I tested successfully

PR #627, opened 2022-06-13
Common: fix mysql connection

oguzhanunlu:
See #626 for details of the bug.

Also, `useSSL` & `verifyServerCertificate` are deprecated in favor of new `sslMode` flag

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
@@ -59,13 +59,13 @@ object Rdbms {
     database: String
   ) extends Rdbms {
 
-    val driver: Class[_] = Class.forName("com.mysql.jdbc.Driver") // Load class
+    val driver: Class[_] = Class.forName("com.mysql.cj.jdbc.Driver") // Load class
 
     val connectionString =
       s"jdbc:mysql://$host:$port/$database?user=$username&password=$password" ++ (if (sslMode)
-                                                                                    "&useSsl=true&verifyServerCertificate=false"
+                                                                                    "&sslMode=REQUIRED"
                                                                                   else
-                                                                                    "")
+                                                                                    "&sslMode=PREFERRED")
   }
 
   val postgresqlDbDecoder: Decoder[PostgresqlDb] =
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -35,7 +35,7 @@ object Dependencies {
     val useragent        = "1.21"
     val uaParser         = "1.4.3"
     val postgresDriver   = "42.2.25"
-    val mysqlConnector   = "8.0.16"
+    val mysqlConnector   = "8.0.29"
     val hikariCP         = "5.0.1"
     val jaywayJsonpath   = "2.7.0"
     val iabClient        = "0.2.0"


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
```diff
@@ -59,13 +59,13 @@ object Rdbms {
...
-    val driver: Class[_] = Class.forName("com.mysql.jdbc.Driver") // Load class
+    val driver: Class[_] = Class.forName("com.mysql.cj.jdbc.Driver") // Load class
 
     val connectionString =
       s"jdbc:mysql://$host:$port/$database?user=$username&password=$password" ++ (if (sslMode)
-                                                                                    "&useSsl=true&verifyServerCertificate=false"
+                                                                                    "&sslMode=PREFERRED"
                                                                                   else
-                                                                                    "")
+                                                                                    "&sslMode=DISABLED")
```
istreeter, line 68: Did you think much about `PREFERRED` and `DISABLED` vs [the other options available](https://dev.mysql.com/doc/refman/5.7/en/connection-options.html#option_general_ssl-mode)?

[The enrichment iglu schema](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/sql_query_enrichment_config/jsonschema/1-0-0) does not explain the meaning of the sslMode field.  But [our documentation says](https://docs.snowplowanalytics.com/docs/enriching-your-data/available-enrichments/custom-sql-enrichment/):

> `sslMode`, whether the database **requires** connections to use SSL

(emphasis mine)

So if it is `true` in the enrichment configuration, I reckon we should use `sslMode=REQUIRED`, not preferred.  It would be nice to use an even higher setting, to also prevent MITM vulnerability, but we can probably assume the DB is within a VPC which we can trust.

If it is `false` in the enrichment configuration, it is probably because the user does not want the CPU overhead of encryption, and because the DB is on a trusted host in a secure VPC.  So `DISABLED` is probably a good choice.

----

Interesting to consider what we're doing with the postgres connection too.

For postgres we use the setting `ssl=true`, which is the same as `ssl=verify-full`.  [It means](https://jdbc.postgresql.org/documentation/head/ssl-client.html):

> the server host name will be verified to make sure it matches the name stored in the server certificate.

So the equivalent in mysql is `sslMode=VERIFY_IDENTITY`.

And when ssl is unset in the connection string, the postgres default is `PREFER`.  It is very interesting to read [what the postgres docs say about this option!](https://www.postgresql.org/docs/current/libpq-ssl.html):

> The default value for sslmode is prefer. As is shown in the table, this makes no sense from a security point of view, and it only promises performance overhead if possible. It is only provided as the default for backward compatibility, and is not recommended in secure deployments.

istreeter: Hey @oguzhanunlu good research!  Regarding MITM attacks -- there are loads of places in the Snowplow pipeline where we use SSL without checking the certificates.  For example, the load balancers forward requests to the collector using HTTPS, but they don't check certificates.  This is a defensible setup, as long as all network traffic is strictly within a trusted VPC.

In other words.... I agree with using REQUIRED, instead of something higher.  It is consistent with the rest of Snowplow.

As for the `false` option -- I would probably have gone with DISABLED, as it allows the user a way to save the cpu overhead, even if the DB server offers SSL.  But on the other hand I respect your decision to keep postgres as it was before, and to align mysql and postgres.  It's a good decision, with strong reasons.

I'm happy to approve this PR, and well done on finding the problem.

oguzhanunlu: Thanks @istreeter , just pushed a separate commit, I'll squash and merge to my remote adapter PR after your approval

PR #630, opened 2022-06-17
Drop KPL / Sink a chunk rather than individual event

benjben:
### Context

[enrich-kinesis](https://github.com/snowplow/enrich/tree/master/modules/kinesis/src) is the replacement for [stream-enrich-kinesis](https://github.com/snowplow/enrich/tree/master/modules/stream/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream).

[common-fs2](https://github.com/snowplow/enrich/tree/master/modules/common-fs2/src) (also used by `enrich-pubsub`) contains all the logic, apart from the sourcing and sinking.

### Changes

This PR aims at changing the way we send records to Kinesis (more details on https://github.com/snowplow/enrich/issues/633), to use [the same mechanism](https://github.com/snowplow/enrich/blob/master/modules/stream/kinesis/src/main/scala/com.snowplowanalytics.snowplow.enrich.stream/sinks/KinesisSink.scala#L201) as `stream-enrich` and [manually build PutRecords requests](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L181-L199) instead of having KPL doing it for us.

The main consequence is that now the sinking functions ([contained in the Environment](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala#L97-L99)) are defined [for a chunk](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L184-L214) whereas before it was defined for [individual records](https://github.com/snowplow/enrich/blob/3.1.5/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L171-L183).

What's important to have in mind is that a `PutRecords` request can be successful but [contain failures](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L139) that need to be retried.

So there are 2 types of retries:
- When the `PutRecords` request [fails](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L116-L126)
- When the successful result [contains failures](https://github.com/snowplow/enrich/blob/fix/kinesis_throughput/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala#L144-L163)


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -168,42 +180,80 @@ object Enrich {
...
     chunk: List[(B, Result)],
-    sinkOne: Validated[BadRow, EnrichedEvent] => F[Unit],
-    trackLatency: Option[Long] => F[Unit]
-  ): F[List[B]] =
-    chunk.parTraverse {
-      case (record, result) =>
-        val (events, collectorTstamp) = result
-        events.parTraverse_(one => sinkOne(one)).as(record) <* trackLatency(collectorTstamp)
-    }
+    env: Environment[F, A]
```
istreeter, line 185: I'm only pointing this out in case you find it interesting:

You never use the type parameter `B`, which means you could drop B from the method signature like this:

```
 def sinkChunk[F[_]: Concurrent: Parallel, A](
    chunk: List[(Any, Result)],
    env: Environment[F, A]
): F[Unit] = {
```

It works because List and Tuple2 are covariant in B.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -168,42 +180,80 @@ object Enrich {
...
+        env.sinkGood,
+        env.metrics.goodCount,
+        env.metadata.observe,
+        env.sinkPii,
+        env.piiAttributes,
+        env.processor,
+        env.streamsSettings.maxRecordSize
+      ),
+      sinkBad(allBad, env.sinkBad, env.metrics.badCount),
+      env.metrics.enrichLatency(chunk.headOption.flatMap(_._2._2))
```
istreeter, line 212: I suggest update the enrichLatency only after writing good.  Otherwise we are lying slightly about the latency of data reaching the enriched stream.

istreeter (commented):
I only reviewed common fs2, like you said.  Looking good though!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -169,42 +181,84 @@ object Enrich {
...
+    attributes: EnrichedEvent => Map[String, String],
+    processor: Processor,
+    maxRecordSize: Int
+  ): F[Unit] =
+    maybeSink match {
+      case Some(sink) =>
+        val (bad, serialized) =
+          enriched
+            .map(ConversionUtils.getPiiEvent(processor, _))
+            .collect { case Some(e) => e }
```
istreeter, line 250: Only sharing this because it's kind fun....

I think this can be `.flatMap(ConversionUtils.getPiiEvent(processor, _))`  instead of `.map(....).collect(....)`

Untested though.

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -14,164 +14,208 @@ package com.snowplowanalytics.snowplow.enrich.kinesis
...
+      failuresRetried <- if (putRecordsResult.getFailedRecordCount != 0) {
+                           val failurePairs = records.zip(putRecordsResult.getRecords.asScala).filter(_._2.getErrorMessage != null)
+                           val (failedRecords, failedResults) = failurePairs.unzip
+                           val logging = logErrorsSummary(getErrorsSummary(failedResults))
+                           val maybeRetrying =
+                             for {
+                               nextRetry <- retryPolicy.decideNextRetry(retryStatus)
+                               maybeRetried <- nextRetry match {
+                                                 case PolicyDecision.DelayAndRetry(delay) =>
+                                                   writeToKinesis(
```
istreeter, line 148: A more elegant solution would be explicitly tail recursive.  It's hurting my brain to work out if this is susceptible to stack overflow exception -- I can't tell.

Luckily the number of retries is small, so we're not going to get stack overflow.  For that reason I'm OK if you don't change anything for now.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: Yes good point !

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
istreeter: Good plan!

istreeter (approved):

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
benjben: https://github.com/snowplow/enrich/issues/644

PR #642, opened 2022-06-23
common-fs2: fix the building of metadata (close #634)

### LINKED ISSUE #634: common-fs2: fix the building of metadata
Because of the `++` [here](https://github.com/snowplow/enrich/blob/3.1.5/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/experimental/Metadata.scala#L244) the values are not concatenated for each key and as a result metadata can be incomplete

benjben:


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/experimental/Metadata.scala
```diff
@@ -241,7 +241,9 @@ object Metadata {
     )
 
   def recalculate(previous: EventsToEntities, events: List[EnrichedEvent]): EventsToEntities =
-    previous ++ events.map(e => MetadataEvent(e) -> unwrapEntities(e))
+    (previous.toList ++ events.map(e => MetadataEvent(e) -> unwrapEntities(e)))
+      .groupBy(_._1)
+      .map { case (k, v) => k -> v.flatMap(_._2).toSet }
```
istreeter, line 246: How about using [cats semigroup syntax](https://typelevel.org/cats/typeclasses/semigroup.html) to do the hard work of merging the map of sets:

```
previous |+| events.map(e => MetadataEvent(e) -> unwrapEntities(e)).toMap
```

benjben: Nice but the `.toMap` is also a problem because it doesn't concatenate values

benjben: I could create a list of `Map` of size 1 though

istreeter (approved):
Whoops!  I missed that, you're right.  Your new solution is very neat.

I guess you could also do:

```
(previous :: events.map(e => Map(MetadataEvent(e) -> unwrapEntities(e)))).combineAll
```

...but it's not much different.

PR #655, opened 2022-07-21
Common: BigDecimal instead of Float for numeric EnrichedEvent fields

istreeter:


modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/blackbox/enrichments/CurrencyConversionEnrichmentTransactionItemSpec.scala
```diff
@@ -47,7 +47,7 @@ class CurrencyConversionEnrichmentTransactionItemSpec extends Specification with
         "ti_sku" -> "PBZ1001",
         "ti_quantity" -> "2",
         "ti_category" -> "APPAREL",
-        "ti_price" -> "2000.0",
+        "ti_price" -> "2000",
```
dilyand, line 50: Just curious about all changes where you've removed the fractional part. Wouldn't it be better if tests use a fraction, to potentially catch cases where BigDecimal works but BigInt doesn't?

dilyand (approved):
👍🏻

Just one minor question about the tests but not a blocker.

PR #658, opened 2022-07-24
Enrich-kinesis: recover from IllegalArgumentException when checkpointing near end of shard

istreeter:


benjben (approved):
:+1:

benjben:
```
I implemented it so we catch the exception only if it's the last record in the shard

case _: IllegalArgumentException if record.isLastInShard =>

But I'm wondering if I should catch all illegal argument exceptions. Because I think if my theory is right, (that KCL internal state can get updated to only allow shard end checkpointing) then this exception could also happen on the penultimate batch for a shard, or earlier.  Because of how we pull a few batches in advance into a queue.
On the one hand, we've only ever seen this exception happen on the final record.  So I'm reluctant to catch and ignore an exception for a situation we've never seen.But on the other hand, if we ever do get an IllegalArgumentException on a non-final record, then there is no point in re-trying the checkpoint.  And it might be better to ignore it and move on, instead of crashing the app.
```

Tough call.

> this exception could also happen on the penultimate batch for a shard, or earlier.  Because of how we pull a few batches in advance into a queue.

I agree that we could possibly end up in a situation where we are stuck in infinite failures. An ideal solution would be to catch the `IllegalArgumentException` and then check if the shard where the record belongs to has been splitted. If yes, ignore the checkpointing error. If not, retry or crash. First I'd like to understand how we can end up in this scenario of inconsistent state.

istreeter:
> catch the `IllegalArgumentException` and then check if the shard where the record belongs to has been splitted

How about....

```
case _: IllegalArgumentException if (record.isLastInShard || record.lastRecordSemaphore.hasQueuedThreads()) =>
```

It's not quite what you said, but it's the best I can come up with using the methods we have available on the record.  The return value of `.hasQueuedThreads()` should only be true if we're at or near the end of a shard.  See lines 37 and 46 of [this file in fs2-aws](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala) for where the record processor acquries the semaphore.

benjben:
> we've only ever seen this exception happen on the final record

How do you know that it was always for the final record of the final batch ? I feel that the exception happened especially when checkpointing for a previous batch, otherwise we would have tried to checkpoint `SHARD_END` and not a sequence number.

> See lines 37 and 46

Line 37 is called [when the shard that this record process is handling has been completed](https://github.com/awslabs/amazon-kinesis-client/blob/a344ee4d05d19d4786a7a4d5d5062adcc080a8c0/amazon-kinesis-client/src/main/java/software/amazon/kinesis/processor/ShardRecordProcessor.java#L58). I hope that this doesn't happen before we manually call `checkpoint`, otherwise we could possibly [checkpoint](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L38) the shard end before processing some pending events (`shardEnded` [is called](https://github.com/awslabs/amazon-kinesis-client/blob/a344ee4d05d19d4786a7a4d5d5062adcc080a8c0/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ShutdownTask.java#L222) in the `ShutdownTask`).

For line 46, do you think that [processRecords](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L44) (and thus `lastRecordSemaphore.acquire()`) is called before the chunk of records is being actually processed by enrich? It's called [here](https://github.com/awslabs/amazon-kinesis-client/blob/a344ee4d05d19d4786a7a4d5d5062adcc080a8c0/amazon-kinesis-client/src/main/java/software/amazon/kinesis/lifecycle/ProcessTask.java#L226) but I need a deeper understanding of KCL.

benjben:
Actually I think that the `Semaphore` we want to use [belongs to one particular `Chunk`](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L26) and can't be used for cross chunks purposes.

istreeter:
I'll explain how I think it works...

When KCL invokes [processRecords](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L44-L72) on the _final_ batch, the sequence number is _not_ `SHARD_END`; it's an actual number representing the last message in the shard.

When KCL invokes [shardEnded](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L35-L39) then the sequence number is `SHARD_END`, not a real number.

KCL expects us to checkpoint the last sequence number (real number) and subsequently the shard end.

So I disagree with....

> I feel that the exception happened especially when checkpointing for a previous batch, otherwise we would have tried to checkpoint SHARD_END and not a sequence number.

All we know is it's happening when the sequence number is a real number.  This could be a non-last batch, but I am sure it could also be the last batch. I agree though that we have no proof which one.

> Line 37 is called when the shard that this record process is handling has been completed. I hope that this doesn't happen before we manually call checkpoint

It does happen before we manually call checkpoint!  But it's nothing to worry about, because of how fs2-aws cleverly uses the semaphore.  [The line](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L37) `lastRecordSemaphore.acquire()` blocks the thread, which means it delays the call to `shardEndedInput.checkpointer().checkpoint()` until the semaphore is released.  The semaphore is only released [when we checkpoint the last record in the shard](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/CommittableRecord.scala#L35).

> For line 46, do you think that processRecords (and thus lastRecordSemaphore.acquire()) is called before the chunk of records is being actually processed by enrich?

Yes, it definitely is called before the chunk of records ibe processed by enrich.  In Enrich language, that function is part of our "source".  It calls `cb(Chunk(chunk: _*))` (line 70) which triggers [this line](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/Kinesis.scala#L86) to enqeueue the records onto the source stream.

> Actually I think that the Semaphore we want to use belongs to one particular Chunk and can't be used for cross chunks purposes.

I think I disagree with this too.  There is one semaphore per `ChunkedRecordProcessor`, and there is one `ChunkedRecordProcessor` per shard.  Each chunked record processor emits many chunks; one chunk for each batch of records it receives from KCL.

benjben (commented):
I agree with everyting you've said!

```
record.lastRecordSemaphore.hasQueuedThreads()
```
Let's say that checkpointing fails for penultimate batch. While we are processing it, final batch has already been queued by `fs2-aws` and [acquire](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L46) has already been called and it successfully got the permit already, so it's not queued. If [shardEnded](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/ChunkedRecordProcessor.scala#L35) has not been called yet, then there is nothing waiting for a permit and `hasQueuedThreads()` would return `false`. I'm not sure if we could be in this scenario but as a security I would add another condition:

```
case _: IllegalArgumentException if (record.isLastInShard || record.lastRecordSemaphore.hasQueuedThreads() || record.lastRecordSemaphore.availablePermits() == 0) =>
```

istreeter:
@benjben please take a look and see what you think.  I took your idea of using `record.lastRecordSemaphore.availalbePermits === 0` because I agree it covers an extra edge case.  But I dropped using `record.lastRecordSempahore.hasQueuedThreads` because I think this is implied by checking the available permits.

The hardest part of this commit was picking the wording of the log message!

benjben (approved):
Looks great :+1: 

I look forward to seeing what appears in the logs.

istreeter:
I did some testing, in which I sent a large number of events through enrich, and scaled the number of kinesis shards up and down.

I saw some log messages like this (i.e the ones we added for not-the-last batch):

```
[pool-1-thread-1] WARN com.snowplowanalytics.snowplow.enrich.kinesis.KinesisRun - Checkpointing failed on a record which was not the last in the shard. Meanwhile, KCL has already enqueued the final record in the shard to the fs2 queue. Ignoring error and instead continue processing towards the shard end
```

But I did not see any log messages with the warning we added for the last batch ([this line](https://github.com/snowplow/enrich/blob/049533e3867013cd87add279a6a373b9b380fbd4/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/KinesisRun.scala#L98)).  My conclusion is that all the errors we saw before were because of **non-final** batches.  It seems KCL is OK with handling checkpointing the final record after reaching the shard end.

So it's possible that we never needed [this block](https://github.com/snowplow/enrich/blob/develop/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/KinesisRun.scala#L93-L100) for recovering from exceptions on the **final** record in the shard.  However, I lean towards keeping it anyway.  Because our fix looks robust, and this whole investigation was a little bit mysterious, so it feels safer to have a robust fix for all possible exceptions.

benjben:
Thanks a lot for the feedback, that's great to see that what we came up with fixed the issue.

I agree with you with keeping the other block as well.

PR #666, opened 2022-08-05
Update Mailgun adapter to work with the new API

peel:
Previously it was impossible to use Mailgun adapter.
Mailgun adapter has now been updated to work with `application/json` webhooks.
These move all the parameters (with a similar schema) from form-encoded into
json object fields. The code has been simplified to reflect on the change.


dilyand (approved):
👍🏻 Looks great @peel! Much neater now too.

Issue #669, opened 2022-08-05
Common: api_request_enrichment_config schema version 1-0-1 support


lmath:
Back in 2018, we discussed adding an ability to disable the cache for Api Request enrichments in [this thread](https://github.com/snowplow/snowplow/issues/3857). Note that the title references POST requests, but as it happens [1-0-0](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-0) already allows POST, and if you diff 1-0-0 vs 1-0-1 the difference is just in whether the min value for cache size is 0 or 1. 

Following that thread, version 1-0-1 of the api_request_enrichment_config schema was added here:

https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-1

However, we did not make the [corresponding change](https://github.com/snowplow/enrich/blob/6c14825f1e98e2d5d1c0e90de4cda5866c44c07f/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala#L44) in enrich to support version 1-0-1. 

It seemed reasonable in 2018 to allow users setting up an Api Request Enrichment to disable the cache. It still seems reasonable to me now, as they will know the most about their API and be best placed to make the call about balance between extra traffic to their API vs accuracy.

PR #670, opened 2022-08-05
Common: api_request_enrichment_config schema version 1-0-1 support

lmath:
See: https://github.com/snowplow/enrich/issues/669

### check 1 - api request enrichment with schema version 1-0-0 works on a pipeline
this change is in 3.2.4-rc2
1. I brought up a GCP pipeline and ensured enrichment version was set to 3.2.4-rc2
2. I set up a minimal api_request_enrichment_config.json which configures an api, one input and one output which is valid for schema version 1-0-0 (has a non-zero value for cache size)
3. I added a schema to the dev sandbox iglu server to configure the output in api_request_enrichment_config.json
4. I added a subscription to the good event topic so that I'd be able to pull and look at good events
5. I sent a test event to the collector of my GCP pipeline
6. I was able to see the good event coming through, enriched with the output field I had defined in api_request_enrichment_config.json and matching the schema I added to the iglu server.

### check 2 - api request enrichment with schema version 1-0-1 works on a pipeline
same as above, except for 2:
I set up a minimal api_request_enrichment_config.json which configures an api, one input and one output which is valid for schema version 1-0-1 (where cache size is 0) and then re-ran `terraform apply`

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
@@ -40,7 +40,6 @@ object ApiRequestEnrichment extends ParseableEnrichment {
       "api_request_enrichment_config",
       "jsonschema",
       1,
-      0,
       0
     )
 


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
```diff
@@ -40,7 +40,6 @@ object ApiRequestEnrichment extends ParseableEnrichment {
       "api_request_enrichment_config",
       "jsonschema",
       1,
-      0,
```
lmath, line 43: I wondered initially if the criteria should be `valid == 1-0-0 or 1-0-1` rather than `valid == 1-0-x`

but it seems like ApiRequestEnrichment was the outlier here. See: [`WeatherEnrichement`](https://github.com/snowplow/enrich/blob/65d7f05e6937393bf983ed91467b5c7dcac3d34d/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/WeatherEnrichment.scala#L49), [`YauaaEnrichement`](https://github.com/snowplow/enrich/blob/481760213049a86c84099c33679293c0af9aef0e/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala#L38), [`HttpHeaderExtractorEnrichment`](https://github.com/snowplow/enrich/blob/65d7f05e6937393bf983ed91467b5c7dcac3d34d/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/HttpHeaderExtractorEnrichment.scala#L33)

dilyand (approved):
Loved the added history and reasoning.👍🏻

PR #674, opened 2022-08-11
enrich-pubsub: Fix to sending output events in parallel

### LINKED ISSUE #673: enrich-pubsub: Fix to sending output events in parallel
Before the 3.2.0 release, we used [`parTraverse` on a batch of events](https://github.com/snowplow/enrich/blob/3.1.5/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala#L179), so that they got sent to pubsub in parallel.  But starting from 3.2.0, we started using [`producer.produceMany`](https://github.com/snowplow/enrich/blob/3.2.3/modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Sink.scala#L63), which defers to the 3rd party lib on how to send the batch of events.  Under the hood, the permutive lib does [a simple `traverse` instead of `parTraverse`](https://github.com/permutive-engineering/fs2-google-pubsub/blob/v0.18.1/fs2-google-pubsub-grpc/src/main/scala/com/permutive/pubsub/producer/grpc/internal/DefaultPublisher.scala#L36), which means we are now waiting for each event to get written to pubsub, before starting to write the next event.

The result is that 3.2.0 is much much slower than previous releases for enrich-pubsub.

istreeter:


dilyand (approved):
👍🏻

Issue #639, opened 2022-06-21
Common: Update Yauaa to 7.x and add support for parsing Client Hints


nielsbasjes:
With version 7.0.0 the Yauaa User-Agent analyzer now has support for also parsing the User-Agent Client Hints. See https://yauaa.basjes.nl/using/clienthints/

In addition to only providing a `User-Agent` value it is now possible to provide a Map<String, String> with the values for the relevant request headers (i.e. headers like `User-Agent` and `Sec-CH-UA-Mobile`) and get a more accurate analysis result (where possible).

istreeter:
Thanks @nielsbasjes!  At first I assumed this meant just bump the library to 7.x and do nothing else.  But I see now there is an opportunity for snowplow to do something bigger and better than that, to take advantage of this new yauaa feature.

Currently, we call `ua.parse(userAgent)` [here](https://github.com/snowplow/enrich/blob/3.3.2/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala#L107) where `userAgent` is a simple string taken from the headers.  But we should change to instead to pass in a map comprising all useragent and client-hint headers [as described in these docs](https://yauaa.basjes.nl/using/clienthints/#analyzing-the-user-agent-client-hints).

I assume we don't need to also make a collector change to request these hints from the browser.  The docs says:

> By default the browsers that support this will send the “low entropy” values without the need to do anything special (other than going over https).

So to start with I think we can do this without a collector change, and still get the benefit of passing in the client hints.  It's interesting to consider if in future we could change the collector to ask for any higher entropy client hint.

paulboocock:
The JS Tracker has a feature to read the Client Hints from the Browser Client Hints API: https://docs.snowplow.io/docs/collecting-data/collecting-from-own-applications/javascript-trackers/browser-tracker/browser-tracker-v3-reference/plugins/client-hints/

We could perhaps leverage this captured schema, although I imagine getting the headers directly in the request will make this more seemless for end users.

istreeter:
@paulboocock I'm veering off topic here, but we could probably write an enrichment to add `org.ietf/http_client_hints` as a derived context, instead of doing it in the tracker.  Looks like it's just a matter of mapping headers to fields.

nielsbasjes:
@istreeter 
Yes this is correct.
I see the following steps:
[...]

PR #678, opened 2022-08-16
Common: Upgrade yauaa to 7.x

oguzhanunlu:
- [ ]  add yauaa_context/1-0-4 schema to commonFs2 test/iglu-client-embedded schemas
- [ ] change com.snowplowanalytics.snowplow.enrich.common.fs2.enrichments.YauaaEnrichmentSpec to use new version


istreeter (commented):
So the original issue (#639) is about more than simply bumping the yauaa library version.  I believe the issue was also requesting us to make a code change to use a new feature of yauaa.  See [my comment in the issue](https://github.com/snowplow/enrich/issues/639#issuecomment-1248290853).

integration-tests/enrich-kafka/config/iglu_resolver.json
```diff
@@ -9,7 +9,7 @@
         "vendorPrefixes": [ "com.snowplowanalytics" ],
         "connection": {
           "http": {
-            "uri": "http://iglucentral.com"
+            "uri": "http://iglucentral-dev.com.s3-website-us-east-1.amazonaws.com/feature/yauaa-1-0-4/"
```
istreeter, line 12: This line will be removed when [the iglu central PR](https://github.com/snowplow/enrich/pull/678) is merged.

modules/common/src/test/resources/iglu-client-embedded/schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
```diff
@@ -0,0 +1,233 @@
+{
```
dilyand, line 1: It seems like we have a duplicate of this schema here and in `common-fs2/src/test`. Can we not use a single file? And do we need it hardcoded once the iglu-central PR is merged?

dilyand (commented):
According to [this comment](https://github.com/snowplow/iglu-central/pull/1238#issuecomment-1288750608), we're supposed to be filtering out a field. But I didn't see any changes to that effect.

In addition, I've added a few comments about things that can be improved. The main one being, that at present, if we have an explicit `null` or `''` useragent, then we ignore all the client hint headers. I just want to ensure that this is indeed what we want to be doing.

project/Dependencies.scala
```diff
@@ -40,7 +40,8 @@ object Dependencies {
     val hikariCP         = "5.0.1"
     val jaywayJsonpath   = "2.7.0"
     val iabClient        = "0.2.0"
-    val yauaa            = "5.23"
+    val yauaa            = "7.7.0"
+    val log4jToSlf4j     = "2.18.0"
```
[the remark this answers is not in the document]
istreeter: Yes, [this is the related issue](https://github.com/nielsbasjes/yauaa/issues/408) in yauaa.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala
```diff
@@ -91,23 +91,91 @@ final case class YauaaEnrichment(cacheSize: Option[Int]) extends Enrichment {
...
+  def getYauaaContext(userAgent: String, headers: List[String]): SelfDescribingData[Json] =
+    SelfDescribingData(YauaaEnrichment.outputSchema, parseUserAgent(userAgent, headers).asJson)
 
   /**
    * Gets the map of attributes retrieved by YAUAA from the user agent.
    * @return Map with all the fields extracted by YAUAA by parsing the user agent.
    *         If the input is null or empty, a map with just the DeviceClass set to Unknown is returned.
    */
-  def parseUserAgent(userAgent: String): Map[String, String] =
+  def parseUserAgent(userAgent: String, headers: List[String]): Map[String, String] =
```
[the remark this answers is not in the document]
istreeter: The `userAgent` field can come from two places:

- Extracted from the header
- Explicitly set as a field in the event.

By the time we reach this enrichment, we have dropped information about where `userAgent` originally came from.  This cannot be easily changed without quite a big re-write of the enrichment logic.

If the user agent is null or empty, then this request did not original from a browser.  So it is fine to ignore client hint headers, because they are only relevant for web.

We cannot change the function to take a single `headers` field.  Imagine the case of server-side tracking where the headers contains the user agent of the server, but the tp2 payload explicitly sets the user agent of the end client.  In this case, we need to override the `User-Agent` header in the list.  I put in a test for this case, so we don't get any accidental regressions if someone tries to change the code.

Renaming the function is a good idea.  How about `analyzeUserAgent` -- afterall that's what the 3rd `a` in yauaa stands for.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/YauaaEnrichment.scala
```diff
@@ -91,23 +91,91 @@ final case class YauaaEnrichment(cacheSize: Option[Int]) extends Enrichment {
...
+        val parsedUA = uaa.parse(headerMap.asJava)
         parsedUA.getAvailableFieldNamesSorted.asScala
           .map(field => decapitalize(field) -> parsedUA.getValue(field))
           .toMap
-          .filter(_._1 != "__SyntaxError__")
+          .filterKeys(validFields)
     }
+
+  /** Yauaa 7.x added many new fields which are not in the 1-0-4 schema */
+  private val validFields = Set(
```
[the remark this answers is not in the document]
istreeter: I thought this too! If you think more widely, it's a bigger piece of functionality that is currently missing from Iglu libs:

Given some json (the output of yauaa enrichment) and given a schema (the yauaa 1-0-4) schema, cast the json so it is valid against the schema.  In this case, drop fields and keep just the valid fields.

I did not implement it here for a few reasons:
1. Yauaa enrichment would then have a dependency on Iglu client, to get the schema.  It makes the code change quite a lot bigger, because it introduces a `F[_]` and we have to start handling failures to fetch the schema.
2. It is [a temporary problem](https://github.com/snowplow/snowplow-rdb-loader/issues/1065) that we cannot upgrade the yauaa schema to add extra fields.  Once that problem is fixed then we don't need to maintain this list.

It's a balance.  Weighing up 1 vs 2, and considering we have users waiting for this feature, I decided to implement it the easy way.

istreeter:
> According to [this comment](https://github.com/snowplow/iglu-central/pull/1238#issuecomment-1288750608), we're supposed to be filtering out a field. But I didn't see any changes to that effect.

We are filtering out the field.  [This is the relevant line](https://github.com/snowplow/enrich/pull/678/commits/0a20eac1bafd82c833f0b071f73a070358edf0ba#diff-284823759c39077cdfd1139f182e876ad7bdb373836b44f0762a42b1f258b706R117).

modules/common/src/test/resources/iglu-client-embedded/schemas/nl.basjes/yauaa_context/jsonschema/1-0-4
istreeter: This is the tip of something bigger.  We have [a lot of iglu central schemas](https://github.com/snowplow/enrich/tree/3.5.0/modules/common-fs2/src/test/resources/iglu-client-embedded/schemas) hardcoded into Enrich's test resources.  Several of them are duplicated between `common` and `commonFs2`.

Yes I could deduplicate those duplicates, but then it would look a bit odd that `commonFs2` has some gaps in the long list of schemas that it maintains.

> do we need it hardcoded once the iglu-central PR is merged?

I like how they are locally available when the tests run.  For unit tests especially, it is good to avoid going to the network.  But that doesn't mean they need to be hardcoded in Enrich!

One solution could be to download selected iglu central schemas before the tests run, and add them to a shared resources directory so they are available for all tests.  It could be implemented as a sbt task, so that `test` depends on this new task.  It would provide a better guarantee that our code is valid against the real Iglu Central.

What do you think?

dilyand: I think that would be great! It exposes us to some rogue iglu central patch; but that seems less likely than us forgetting to update a schema somewhere in the hardcoded files.

istreeter: Idea for the future, but I will create an issue now.

dilyand (approved):
👍🏻

istreeter:
Will merge as soon as the Iglu Central schema is released.

PR #680, opened 2022-08-24
Add Flow Control to PubSub consumer

oguzhanunlu:
- [x] Reference value is to be updated after benchmarking

### DIFF
--- a/config/config.pubsub.extended.hocon
+++ b/config/config.pubsub.extended.hocon
@@ -15,8 +15,20 @@
     # of events in memory at once. An event counts towards this limit starting from when it received
     # by the permutive library, until we ack it (after publishing to output). The value must be large
     # enough that it does not cause the sink to block whilst it is waiting for a batch to be
-    # completed.
+    # completed. The first of maxQueueSize and maxRequestBytes being reached will pause the consumption.
     "maxQueueSize": 3000
+
+    # Optional. Configures the "maximum outstanding request bytes" of PubSub subscriber.
+    # It puts an upper bound on the events' bytes that can be hold in memory at once before getting acked.
+    # The value must be large enough to not cause the sink to block whilst it is waiting for a batch to be
+    # completed.
+    # The first of maxQueueSize and maxRequestBytes being reached will pause the consumption.
+    # Default: 50 MB
+    "maxRequestBytes": 50000000
+
+    # Optional. Configures the maximum period a message ack deadline can be extended
+    # A zero duration disables auto deadline extension
+    "maxAckExtensionPeriod": 1 hour
   }
 
   "output": {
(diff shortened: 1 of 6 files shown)


modules/pubsub/src/main/resources/application.conf
```diff
@@ -3,6 +3,8 @@
     "type": "PubSub"
     "parallelPullCount": 1
     "maxQueueSize": 3000
```
benjben, line 5: I would set that much higher like 10k or maybe even 100k, given that we have the bytes limit

modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Source.scala
```diff
@@ -44,12 +47,24 @@ object Source {
...
+        .setMaxOutstandingRequestBytes(input.maxRequestBytes)
+        .setLimitExceededBehavior(LimitExceededBehavior.Block)
+        .build()
+
     val pubSubConfig =
       PubsubGoogleConsumerConfig(
         onFailedTerminate = onFailedTerminate,
         parallelPullCount = input.parallelPullCount,
-        maxQueueSize = input.maxQueueSize
+        maxQueueSize = input.maxQueueSize,
```
benjben, line 63: Do we still need this line?

benjben (approved):
:+1:

modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Source.scala
oguzhanunlu: yes, to set size of fs2-pubsub's internal queue

modules/pubsub/src/main/resources/application.conf
oguzhanunlu: going for 10K after the test against 1KB events

config/config.pubsub.extended.hocon
```diff
@@ -17,6 +17,16 @@
     # enough that it does not cause the sink to block whilst it is waiting for a batch to be
     # completed.
     "maxQueueSize": 3000
```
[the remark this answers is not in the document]
oguzhanunlu: right, thanks Ben!

modules/pubsub/src/main/scala/com/snowplowanalytics/snowplow/enrich/pubsub/Source.scala
benjben: Ah yes I now get the difference, thanks!

istreeter (approved):

modules/pubsub/src/main/resources/application.conf
oguzhanunlu: decided to stay with 3K after observing memory usage

PR #686, opened 2022-08-31
Add integration test of enrich-kafka

oguzhanunlu:
tested on localhost successfully

collector payload generation is a copy-paste from `enrich-kinesis/it`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,13 @@ jobs:
       run: SBT_OPTS="-Xms1G -Xmx8G -Xss4M -XX:MaxMetaspaceSize=1024M" sbt coverage +test
       env:
         OER_KEY: ${{ secrets.OER_KEY }}
+    - name: Run integration tests
+      id: integrationTest
+      run: |
+        sbt "project kafka" "docker:publishLocal"
+        docker-compose -f integration-tests/enrich-kafka/docker-compose.yml up -d
+        sbt "project kafka" "it:test"
+        docker-compose -f integration-tests/enrich-kafka/docker-compose.yml down
     - name: Check Scala formatting
       if: ${{ always() }}
       run: sbt scalafmtCheckAll
--- a/build.sbt
+++ b/build.sbt
@@ -155,8 +155,14 @@ lazy val kafka = project
   .in(file("modules/kafka"))
   .enablePlugins(BuildInfoPlugin, JavaAppPackaging, DockerPlugin)
   .settings(kafkaBuildSettings)
-  .settings(libraryDependencies ++= kafkaDependencies)
+  .settings(libraryDependencies ++= kafkaDependencies ++ Seq(
+    // integration test dependencies
+    specs2CEIt,
+    specs2ScalacheckIt
+  ))
   .settings(excludeDependencies ++= exclusions)
+  .settings(Defaults.itSettings)
+  .configs(IntegrationTest)
   .settings(addCompilerPlugin(betterMonadicFor))
   .dependsOn(commonFs2)
 
--- a/integration-tests/enrich-kafka/config/enrich-kafka.hocon
+++ b/integration-tests/enrich-kafka/config/enrich-kafka.hocon
@@ -1,0 +1,58 @@
+{
+  "input": {
+    "type": "Kafka"
+    "topicName": "it-enrich-kinesis-collector-payloads"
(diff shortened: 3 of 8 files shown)


oguzhanunlu:
I'll try to update gh actions to run kafka with currently tagged enrich-kafka in a docker-compose to make sure each enrich-kafka release passes integration test

oguzhanunlu:
Updated PR based on [this reference](https://github.com/docker/build-push-action/blob/master/docs/advanced/test-before-push.md) which builds images without pushing, to be able to test them before push.

`enrich` service in docker-compose uses current tag to make sure we are running the expected version

I only added kafka:it for now

oguzhanunlu:
moved integration tests upstream in CI workflow, so that it runs for each push

modules/kafka/src/it/scala/com/snowplowanalytics/snowplow/enrich/kafka/test/KafkaSink.scala
```diff
@@ -0,0 +1,37 @@
...
+import cats.effect.{ConcurrentEffect, ContextShift, Resource, Timer}
+import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.Output
+import fs2.kafka.{KafkaProducer, ProducerSettings}
+import fs2.{Pipe, Stream}
+
+import java.util.UUID
+
+object KafkaSink {
+
+  def init[F[_]: ConcurrentEffect: ContextShift: Parallel: Timer](topicName: String, bootstrapServers: String): Pipe[F, Array[Byte], Unit] = {
```
benjben, line 14: Instead of redefining a Kafka sink it would be great to re use [the one that you already have](https://github.com/snowplow/enrich/blob/enrich-kafka/modules/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.kafka/Sink.scala#L27).

I did that in [enrich kinesis integration test](https://github.com/snowplow/enrich/blob/enrich-kafka/modules/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.kafka/Sink.scala#L27)

modules/kafka/src/it/scala/com/snowplowanalytics/snowplow/enrich/kafka/test/KafkaSource.scala
```diff
@@ -0,0 +1,27 @@
+package com.snowplowanalytics.snowplow.enrich.kafka.test
+
+import cats.effect.{ConcurrentEffect, ContextShift, Timer}
+import fs2.Stream
+import fs2.kafka.ConsumerSettings
+import fs2.kafka.KafkaConsumer
+
+object KafkaSource {
+
+  def init[F[_]: ConcurrentEffect: ContextShift: Timer](topicName: String, bootstrapServers: String): Stream[F, Array[Byte]] = {
```
benjben, line 10: Same remark as for the sink

modules/kafka/src/it/scala/com/snowplowanalytics/snowplow/enrich/kafka/test/KafkaSink.scala
benjben: > could you explain more please?

What you need for the integration tests is something that can insert an `Array[Byte]` into Kafka. This is exactly what [this](https://github.com/snowplow/enrich/blob/enrich-kafka/modules/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.kafka/Sink.scala#L30) gives you.

It's great to reuse it for 2 reasons :

- Re use existing pieces instead of reinventing the wheel
- The more we use this code the more chances we have to find bugs

> if we have a bug in main impl, aren't we going to get tricked by successful integration test?

If we have a bug in main impl, integration tests won't be successful

oguzhanunlu:
hey @benjben ,I addressed your feedback in a separate commit, could you take a look please? Thanks in advance

benjben (approved):
Looks great :+1: 

FYI the testing squad has agreed on using [testcontainers](https://www.testcontainers.org/) to run Docker containers for integration tests (like [here](https://github.com/snowplow/stream-collector/pull/258/files#diff-ae458aa4902052d07bfd45953cb94b8eff154adbabc4e6d70757dde66a2b1bf7R35)) so that the whole test can be run with just `sbt 'project integrationTests' test`.

If you're happy to try it, you can update your PR. Otherwise I can take care of it later to not delay this release.

PR #687, opened 2022-09-05
enrich-kinesis: never send batch of records > 5 MB with putRecords

benjben:


modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -105,11 +104,50 @@ object Sink {
...
+    */
+  private[kinesis] def group[A](
+    records: List[A],
+    recordLimit: Int,
+    sizeLimit: Int,
+    getRecordSize: A => Int
+  ): List[List[A]] = {
+    case class Batch(size: Int, count: Int, records: List[A])
+
+    records.foldLeft(List(Batch(0, 0, Nil))){ case (acc, record) =>
```
istreeter, line 129: Would it be safer to do `foldLeft(List.empty[Batch]) {` in case `records` is an empty list?

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -105,11 +104,50 @@ object Sink {
...
+      acc match {
+        case head :: tail =>
+          if (head.count + 1 > recordLimit || head.size + getRecordSize(record) > sizeLimit)
+            List(Batch(getRecordSize(record), 1, List(record))) ++ List(head) ++ tail
```
istreeter, line 133: Nit: You call `getRecordSize(record)` twice, whereas you could call it once after line 131.

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
```diff
@@ -39,6 +39,9 @@ import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.Output
 
 object Sink {
 
+  // PutRecords request can be up to 5 MiB, including partition keys
+  val putRecordsSizeLimit = 5 * 1024 * 1024
```
istreeter, line 43: There are a few places in our snowplow estate where a config option has a _theoretical_ maximum, but we set the value slightly lower just in case there is an edge case we don't know about.

- [Here in enrich-pubsub](https://github.com/snowplow/enrich/blob/3.3.0/config/config.pubsub.extended.hocon#L56-L59)
- [Here in collector](https://github.com/snowplow/stream-collector/blob/master/kinesis/src/main/resources/application.conf#L19)

I'm sure there are other places too.  By putting it in `application.conf` it means we can override it in an emergency, just in case we discover an edge case in production.

istreeter (commented):
Looking good Ben :)

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
benjben: I doubt that `records` can be empty but good point, better to be safe than sorry!

benjben:
Here is what I observed:

1. The 1 MB limit for one record in `putRecords` is only for the data, that does not include the partition key. This works:

```
  val oneMB = 1024 * 1024
  val partitionKeyLength = 256

  val partitionKey = "a".repeat(partitionKeyLength)
  val data = "a".repeat(oneMB)
  
  val records = List((partitionKey, data))

  val producer = 
    AmazonKinesisClientBuilder.standard
      .withRegion("eu-central-1")
      .build()

  putRecords(producer, records)

  private def putRecords(
    kinesis: AmazonKinesis,
    records: List[(String, String)]
  ): PutRecordsResult = {
    val entries =
      records.map { case (partitionKey, data) =>
        val prre = new PutRecordsRequestEntry()
        prre.setPartitionKey(partitionKey)
        prre.setData(ByteBuffer.wrap(data.getBytes))
        prre
      }
    val putRecordsRequest = {
      val prr = new PutRecordsRequest()
      prr.setStreamName("ben-testing")
      prr.setRecords(entries.asJava)
      prr
    }
    kinesis.putRecords(putRecordsRequest)
  }
```

Whereas with:
```
val data = "a".repeat(oneMB) + "a"
```
this fails with:
```
[error]   CAUSED BY com.amazonaws.services.kinesis.model.ValidationException: 1 validation error detected: Value at 'records.1.member.data' failed to satisfy constraint: Member must have length less than or equal to 1048576 (Service: AmazonKinesis; Status Code: 400; Error Code: ValidationException; Request ID: d0f000e9-6230-b755-8b60-32c75a889688; Proxy: null)
```

2. The 5 MB limit for the whole `putRecords` **does** include partition key + data. This : 

```
  val oneMB = 1024 * 1024
  val partitionKeyLength = 256

  val partitionKey = "a".repeat(partitionKeyLength)
  val data = "a".repeat(oneMB)
  
  val records = List.fill(5)((partitionKey, data))

  val producer = 
    AmazonKinesisClientBuilder.standard
      .withRegion("eu-central-1")
      .build()

  putRecords(producer, records)
```
fails with:
```
[error]   CAUSED BY com.amazonaws.services.kinesis.model.InvalidArgumentException: Records size exceeds 5 MB limit (Service: AmazonKinesis; Status Code: 400; Error Code: InvalidArgumentException; Request ID: d1cc441d-24a3-0d49-8a5c-7770cd65f739; Proxy: null)
```
But this :
```
  val data = "a".repeat(oneMB - partitionKeyLength)
```
works.

So there is no overhead and the limit is really the sum of all the partition keys' length + all the data length, and we can use `5 * 1024 * 1024 = 5242880`

benjben:
Confirmed with different sizes of partition keys and data in the same `putRecords` : 

```
val dataLength = 1024 * 1024 / 16
  val partitionKeyLength = 256

  val partitionKey = "a".repeat(partitionKeyLength)
  val data = "a".repeat(dataLength - partitionKeyLength)
  
  val fourMB = List.fill(4 * 16)((partitionKey, data))

  val dataLengthBis = 1024 * 1024 / 16 / 16
  val partitionKeyLengthBis = 128

  val partitionKeyBis = "a".repeat(partitionKeyLengthBis)
  val dataBis = "a".repeat(dataLengthBis - partitionKeyLengthBis)
  
  val oneMB = List.fill(1 * 16 * 16)((partitionKeyBis, dataBis))

  val records = fourMB ++ oneMB

  val producer = 
    AmazonKinesisClientBuilder.standard
      .withRegion("eu-central-1")
      .build()

  putRecords(producer, records)
```
works whereas:
```
  val records = fourMB ++ oneMB ++ List(("a","b"))
```
doesn't.

benjben:
@istreeter after these tests do you still think that 5 MB should be made configurable ?

istreeter:
Hey @benjben thanks a lot for the analysis - this gives me confidence that 4194304 is a good safe number.

> do you still think that 5 MB should be made configurable ?

I like the trend over the last couple of years that we have put more and more constants into the `application.conf` file, which is the same thing as saying make it configurable.  I like it there for two reasons:

1. I immediately know which file to look in to find a runtime default, instead of needing to search through the code whenever I'm debugging something
2. Many times we have surprised by something in production and needed to tweak a value.

These days, I find it easier to have a personal rule "make everything configurable" rather than asking myself the question each time "does this need to be configurable".  I don't think I've ever regretted putting something into the `application.conf` file.

I think the way you've done it in your latest commit is perfect, as long as you're also happy with it?

istreeter (approved):

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Sink.scala
istreeter: > I doubt that records can be empty

Actually, I think it can be:

1. If this is the good sink, and all the records in the batch were bad
2. If this is the bad sink, and all the records in the batch were good

But your latest version looks safe.

benjben:
Hey @istreeter ,

> this gives me confidence that 4194304 is a good safe number

Yes it is, and I'm even confident that 5242880 is a safe number, so I'd rather use that one as a default to possibly have a better throughput. Because we will have it configurable we take no risk. WDYT?

istreeter:
I agree!

PR #690, opened 2022-09-08
enrich-kinesis integration test

benjben:
There was already an integration test existing for `enrich-kinesis`, but it was not run by Github Actions, because it needed to create real Kinesis streams in AWS. It now uses `localstack` and `testcontainers`.

There are now 3 tests :

1. [One](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L59-L151) that generates and sends collector payloads to the input Kinesis streams of Enrich in `localstack` and checks that the output streams contain the expected number of enriched events and bad rows. This is the one that was existing before. A big difference is that it's now using the `Source` and the `Sink` of the main module instead of rewriting them ([here](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L91) [here](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L96) and [here](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L79)).
2. [One](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L44-L57) that checks that Enrich can initialize with [the minimal config example](https://github.com/snowplow/enrich/blob/3.5.0/config/config.kinesis.minimal.hocon).
3. [One](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala#L153-L171) that checks that Enrich exits gracefully when it receives a `SIGTERM`.

For `testcontainers`, there is only [one instance](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala#L103) of `localstack` started for [all enrich instances](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala#L104-L106). Communication between Enrich and `localstack` is made possible thanks to using [the same network](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala#L40-L48).

Before running the integration tests, `sbt` [publishes](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/build.sbt#L143) `enrich-kinesis` Docker image locally.


.github/workflows/ci.yml
```diff
@@ -34,34 +34,33 @@ jobs:
...
       run: sbt scalafmtCheckAll
     - name: Check assets can be published
-      if: ${{ always() }}
       run: sbt publishLocal
+    - name: Run integration tests for enrich-kinesis
+      run: sbt "project kinesis" IntegrationTest/test
+    - name: Run integration tests for enrich-kafka
+      run: |
+        sbt "project kafka" "docker:publishLocal"
+        docker-compose -f integration-tests/enrich-kafka/docker-compose.yml up -d
```
colmsnowplow, line 61: Just curious - why do we have docker-compose for kafka but not the other tests?

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
```diff
@@ -0,0 +1,114 @@
...
+import org.testcontainers.containers.wait.strategy.Wait
+import org.testcontainers.containers.output.Slf4jLogConsumer
+
+import com.dimafeng.testcontainers.{FixedHostPortGenericContainer, GenericContainer}
+
+object Containers {
+
+  val (localstack, localtackNetwork) = {
+    val container = FixedHostPortGenericContainer(
+      imageName = "localstack/localstack-light:1.0.4",
```
colmsnowplow, line 27: [Latest is v1.2](https://github.com/localstack/localstack/releases/tag/v1.2.0). Not sure if we'll get much benefit from bumping but if it doesn't break anything might as well.

colmsnowplow (approved):
LGTM! But I think it should also have approval from someone more confident in Scala!

pondzix (commented):
Love it, testcontainers :heart:

modules/kinesis/src/it/resources/localstack/01-createStreams.sh
```diff
@@ -0,0 +1,12 @@
...
+export AWS_ACCESS_KEY_ID=foo
+export AWS_SECRET_ACCESS_KEY=bar
+
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name raw-count --shard-count 1 --region eu-central-1
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name enriched-count --shard-count 1 --region eu-central-1
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name bad-1-count --shard-count 1 --region eu-central-1
+
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name raw-stop --shard-count 1 --region eu-central-1
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name enriched-stop --shard-count 1 --region eu-central-1
+aws --endpoint-url=http://localhost:4566 kinesis create-stream --stream-name bad-1-stop --shard-count 1 --region eu-central-1
```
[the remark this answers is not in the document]
benjben: With the existing tests it would be fine to have only 1 stream of each. If we use a different app name in `count` and `stop` they should consume the streams independently (different KCL DynamoDB tables). I'll update it.

> after each case

The tests are to be run in parallel to save time.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
```diff
@@ -0,0 +1,114 @@
...
+      fileSystemBind = Seq(
+        GenericContainer.FileSystemBind(
+          "modules/kinesis/src/it/resources/localstack",
+          "/docker-entrypoint-initaws.d",
+          BindMode.READ_ONLY
+        )
+      ),
+      waitStrategy = Wait.forLogMessage(".*AWS kinesis.CreateStream.*", 6),
+      exposedHostPort = 4566,
+      exposedContainerPort = 4566
```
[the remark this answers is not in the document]
benjben: This is a leftover from previous implementation without the custom network. I wonder if all the containers of a same network can access each other's ports, I'll try!

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+        for {
+          _ <- isRunning.retryingOnFailures(
+            _ == false,
+            retryPolicy,
+            (_, _) => IO(())
+          )
+        } yield {
+          Containers.enrichMinimal.getLogs() must contain("Parsed config file")
+        }
+      result.unsafeRunSync()
```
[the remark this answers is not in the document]
benjben: There is a [timeout of 10 seconds](https://github.com/typelevel/cats-effect-testing/blob/series/1.x/specs2/shared/src/main/scala/cats/effect/testing/specs2/CatsEffect.scala#L30), which is too short

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
```diff
@@ -0,0 +1,114 @@
...
+    val logs = new Slf4jLogConsumer(logger)
+    container.start()
+    container.followOutput(logs)
+  }
+
+  def startAll(): Unit = {
+    localstack.start()
+    startContainerWithLogs(enrichCount, "count")
+    startContainerWithLogs(enrichStop, "stop")
+    startContainerWithLogs(enrichMinimal, "minimal")
```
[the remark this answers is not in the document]
benjben: Yeah good point. I'd like to have only one `localstack` instance shared by all tests, so I was thinking to start it before all the tests, but `stop` test doesn't need it, so I need to improve that. And you're right it's better to do as you say for enrich instances. I'll update.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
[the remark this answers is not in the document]
benjben: Now I remember why the `FixedHostPortGenericContainer`. We don't need it for Enrich and localstack to communicate. But we do need it for [Kinesis producer and consumer](https://github.com/snowplow/enrich/blob/feature/integration_test_kinesis/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/ConfigCount.scala#L29-L31) that sends and reads data for the integration test `count`.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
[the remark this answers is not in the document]
pondzix: Which... can be overridden? ;)

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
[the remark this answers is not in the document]
pondzix: Yes, but with non-fixed version you can still call `container.getMappedPort(4566)` to get randomly assigned port and use it to configure producer/consumer dynamically in tests instead of hardcoded value?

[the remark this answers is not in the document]
benjben: Ah I now get what you said in your first comment! Good idea, it should work!

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,172 @@
...
+
+  def afterAll: Unit = Containers.stopAll()
+
+  "enrich-kinesis" should {
+    "initialize with the minimal config" in {
+      val isRunning = IO(Containers.enrichMinimal.isRunning())
+      val result =
+        for {
+          _ <- isRunning.retryingOnFailures(
+            _ == false,
```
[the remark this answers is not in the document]
benjben: I dropped the `WaitStrategy` for Enrich containers because it turns out that we don't need it.

For `stop` I think that we can't do without the retry with timeout, because we especially want to check that the container stops and doesn't get stuck, whatever the logs contain.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+      for {
+        enrich <- IO(Containers.enrichMinimal)
+        isRunning = IO(enrich.isRunning())
+        _ <- isRunning.retryingOnFailures(
+          _ == false,
+          retryPolicy,
+          (_, _) => IO(())
+        )
+      } yield {
+        enrich.getLogs() must contain("Running Enrich")
```
pondzix, line 54: Should we call here stop container somewhere?

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  def afterAll: Unit = Containers.localstack.stop()
+
+  "enrich-kinesis" should {
+    "initialize with the minimal config" in {
+      for {
+        enrich <- IO(Containers.enrichMinimal)
+        isRunning = IO(enrich.isRunning())
+        _ <- isRunning.retryingOnFailures(
+          _ == false,
```
pondzix, line 49: So stop retrying when container... is not running? ;) Or I read that wrong.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+        isRunning = IO(enrich.isRunning())
+        _ <- IO(println("stop - Sending signal"))
+        _ <- IO(enrich.getDockerClient().killContainerCmd(enrich.getContainerId()).withSignal("TERM").exec())
+        _ <- isRunning.retryingOnFailures(
+          _ == false,
+          retryPolicy,
+          (_, _) => IO(())
+        )
+        logs = enrich.getLogs()
+        _ <- IO(enrich.stop())
```
pondzix, line 166: Do we need to call `stop` here? `TERM` signal should stop container without explicit call here, is that correct?

modules/kinesis/src/it/resources/enrich/stop.hocon
```diff
@@ -0,0 +1,40 @@
+{
+  "input": {
+    "type": "Kinesis"
+    "appName": "enrich-kinesis-it-stop"
```
pondzix, line 4: I was wondering how adding new tests cases in the future would look like. If I want to add different scenario for my instance of enrich I need to make sure `appName` is unique across all tests? If that's true... how painful is that? 

What options do we have: 

1. New localstack + enrich per test. Everything is isolated, we don't have to worry about anything. Problem: it can take a lot of time. Out of curiosity, how much time does it take to run localstack container? And enrich? 
2. Shared localstack + enrich per test. We need to remember about unique app name. 

What about...generating `appName` dynamically like: `enrich-${UUID}` and setting it using JVM props like: `-D.input.appName` in `GenericContainer` ? Does it make sense or is it overkill?

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+      type AggregateGood = List[Event]
+      type AggregateBad = List[String]
+      case class Aggregates(good: AggregateGood, bad: AggregateBad)
+
+      /** This function is in charge of:
+        * - Generating and sending collector payloads to the input stream of Enrich
+        * - Reading the enriched stream and bad stream populated by Enrich
+        * - Aggregating the above
+        */
+      def run(
```
pondzix, line 68: What do you think about slightly different version, I was thinking about sth like:

```scala
sealed trait OutputRow
object OutputRow {

  final case class Good(event: Event) extends OutputRow
  final case class Bad(line: String) extends OutputRow
}

final case class TestResources(rawSink: ByteSink[IO],
                               goodSource: Stream[IO, OutputRow.Good],
                               badSource: Stream[IO, OutputRow.Bad]

"TEST" in {
        createResources().use { resources =>
          val writingInput = generateInput(resources.rawSink, nbGood = 100, nbBad = 10)
          
          val readingOutput = 
            resources.goodSource
              .merge(resources.badSource)
              .interruptAfter(2.minutes)

          val output = readingOutput.concurrently(writingInput)

          val result = output.compile.toList
          
          //assert result list
        }
} 

  
  private def createResources(): Resource[IO, TestResources] = {
    for {
      blocker <- Blocker[IO]
      rawSink <- Sink.init[IO](blocker, rawConfig) 
    } yield {
      val goodSource = asGood(sourceStream(blocker, enrichedConfiged))
      val badSource = asBad(sourceStream(blocker, badConfig))
      TestResources(rawSink, goodSource, badSource)
    }
  }
  
  private def generateInput(sink: ByteSink[IO], nbGood: Long, nbBad: Long): Stream[IO, Unit] = {
    CollectorPayloadGen.generate[IO](nbGood, nbBad)
      .evalMap(events => sink(List(events)))
  }

  private def sourceStream(blocker: Blocker, config: Input.Kinesis): Stream[IO, Array[Byte]] =
    Source.init[IO](blocker, config, ConfigCount.monitoring)
      .map(KinesisRun.getPayload)
      
  private def asGood(source: Stream[IO, Array[Byte]]): Stream[IO, OutputRow.Good] = {
    source.map { bytes => 
      OutputRow.Good {
        Event.parse(new String(bytes))
          .getOrElse(throw new RuntimeException("can't parse enriched event"))
      }
    }
  }
  private def asBad(source: Stream[IO, Array[Byte]]): Stream[IO, OutputRow.Bad] =
   source.map(bytes => OutputRow.Bad(new String(bytes)))
```

I usually try to make test content as short as possible, but with clear input and output. All test helpers extracted somewhere else, checking impl details only when needed.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/ConfigCount.scala
```diff
@@ -0,0 +1,78 @@
...
+import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.{BackoffPolicy, Input, MetricsReporters, Monitoring, Output}
+
+object ConfigCount {
+
+  val region = "eu-central-1"
+  val appName = "enrich-kinesis-it"
+  val rawStream = "raw"
+  val enrichedStream = "enriched"
+  val badStream = "bad-1"
+  val endpoint = "localhost"
```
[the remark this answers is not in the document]
benjben: Good catch, it was meant to be used [here](https://github.com/snowplow/enrich/pull/690/files#diff-d69c9dfcfcf42e44680dfd6fcd1f8744b47da7445c7ce9a226e21e35276bf2feR77)

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/ConfigCount.scala
```diff
@@ -0,0 +1,78 @@
...
+ */
+package com.snowplowanalytics.snowplow.enrich.kinesis
+
+import java.net.URI
+
+import scala.concurrent.duration._
+
+import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.{BackoffPolicy, Input, MetricsReporters, Monitoring, Output}
+
+object ConfigCount {
```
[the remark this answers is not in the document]
benjben: Ok let's go for TSPCC :laughing:

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala
```diff
@@ -0,0 +1,102 @@
...
+    (container, network)
+  }
+
+  def localStackPort = localstack.container.getMappedPort(4566)
+
+  def enrichCount = enrich("modules/kinesis/src/it/resources/enrich/count.hocon", "count", Some(localtackNetwork))
+  def enrichStop = enrich("modules/kinesis/src/it/resources/enrich/stop.hocon", "stop", Some(localtackNetwork))
+  def enrichMinimal = enrich("config/config.kinesis.minimal.hocon", "minimal")
+
+  def enrich(configPath: String, loggerName: String, network: Option[Network] = None): JGenericContainer[_] = {
```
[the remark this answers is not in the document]
benjben: > What do you think about using this method directly in tests? So when looking at a test, we can see what config is used.

:+1: 

> And why there is no network for minimal case?

Minimal case does not need to communicate with localstack, its only purpose is to check that the config file can be parsed.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+
+  override protected val Timeout = 10.minutes
+
+  val retryPolicy =
+    RetryPolicies.limitRetriesByCumulativeDelay(5.minutes, RetryPolicies.capDelay[IO](2.second, RetryPolicies.fullJitter[IO](1.second)))
+
+  def afterAll: Unit = Containers.localstack.stop()
+
+  "enrich-kinesis" should {
+    "initialize with the minimal config" in {
```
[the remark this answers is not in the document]
benjben: > is the config parsed correctly?

Just this. The reason why we can't test that it fully initializes properly and that it's waiting for data is that communicating with localstack requires to override the default endpoints, but we don't want to do that as it would not be running the minimal file any more. The day we can communicate with real Kinesis in the integration tests, we can go until waiting for data.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: Yeah why not. It's guaranteed that Enrich will crash but it's not guaranteed that it will exit (although it should) ! I'll add this test as well

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,173 @@
...
+      for {
+        enrich <- IO(Containers.enrichMinimal)
+        isRunning = IO(enrich.isRunning())
```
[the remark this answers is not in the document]
benjben: > When does isRunning return true?

When the container is running, so right after `start()`.

Here we don't care when the container is running or not. The only thing that we care about is to see `Running Enrich` in the logs, which means that the configuration has been successfully parsed.

What is probably misleading is that we wait for the container to be stopped (`isRunning = false`, that's what the retries do), and then when it's stopped we check that the logs contain `Running Enrich`. I agree that it would be more clear to have a `WaitStrategy` that waits for this `Running Enrich` log, and right away check again that the logs contain it in the test, because this test isn't about checking that Enrich exits. i'll update.

> And it can be also used for SIGTERM case instead of sleeping

Yes. Although for SIGTERM there is no other choice than having a retry that checks that Enrich exits after receiving the SIGTERM.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: That's correct ! The retry is not here to check that Enrich is now running, we know that it is. It's here to check that it has stopped running.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: Indeed :man_facepalming:

`enrich.stop()` has to be placed after the test `enrich.isRunning() === false`

It should be useless, it's just a security in case Enrich didn't stop after the SIGTERM

benjben: It can't be placed after the test actually, because if the test fails it won't be reached :smile: I need to get `isRunning()` in a variable before `_ <- IO(enrich.stop())` and then use it in the test.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: Yeah I also don't like this very long test, I'll update

modules/kinesis/src/it/resources/enrich/stop.hocon
benjben: > What about... generating each appName randomly ?

Although feasible, I don't think that this is worth the effort, as it requires to template the config files, and update them with the random UUID before mounting them in the container.

I think that putting the test name in the app name is straightforward and good enough. We can revisit that if we suddenly start to add many tests (which will not happen in some near future).

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
pondzix: Maybe we can utilize `Resource` from cats to wrap container start/stop?

modules/kinesis/src/it/resources/enrich/stop.hocon
pondzix: And this: 

>setting it using JVM props like: -D.input.appName

So no appName in the config file, it would be passed by JVM props. No need to template anything then I think?

benjben: I don't get it, having a random UUID is only useful if we have one random UUID *per appName*, how having one for all the tests would help ?

pondzix: I mean set JVM props here: 

```
    val container = GenericContainer(
      dockerImage = "snowplow/snowplow-enrich-kinesis:latest",
      env =
        Map(
          "AWS_REGION" -> "eu-central-1",
          "AWS_ACCESS_KEY_ID" -> "foo",
          "AWS_SECRET_ACCESS_KEY" -> "bar",
         "JVM_HOCON_CONFIG_HERE"
        ),

```

when creating enrich container. Would that work?

pondzix: Or...delete DynamoDB tables after each test case?

benjben: > Or...delete DynamoDB tables after each test case?

No, as we want the tests to run in parallel to save time.

> I mean set JVM props here

Ah right, I get it! It might be possible, let me try.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+  override protected val Timeout = 10.minutes
+
+  val retryPolicy =
+    RetryPolicies.limitRetriesByCumulativeDelay(5.minutes, RetryPolicies.capDelay[IO](2.second, RetryPolicies.fullJitter[IO](1.second)))
+
+  def afterAll: Unit = Containers.localstack.stop()
+
+  "enrich-kinesis" should {
+    "initialize with the minimal config" in {
+      Resource.make[IO, JGenericContainer[_]](
```
pondzix, line 42: Can we hide `Resource` stuff in the `enrich` method? So `enrich` could return `Resource[IO, Container]`? 

Then in a test: 

```scala

"TEST" in {

gimmeMeEnrich(
     config = "This config",
     app/loggerName = "Enrich is named like that in logs"
).use { instance => 
    // do something
}

File: modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala [hunk 19]
```diff
@@ -0,0 +1,110 @@
...
+
+      val nbGood = 100l
+      val nbBad = 10l
+
+      Resource.make[IO, JGenericContainer[_]](
+        IO(
+          Containers.enrich(
+            configPath = "modules/kinesis/src/it/resources/enrich/count.hocon",
+            loggerName = "count",
+            network = Some(Containers.localstackNetwork)
```
pondzix, line 62: Thinking about this `Option[Network]` - maybe we could get rid of the optional and use this network all the time for each enrich container? That would be a detail hidden in the `enrich` method, not leaking to any of a tests. Would that be a problem if we use it for minimal case, even though it doesn't need that? I think it would remove some noise from tests.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+          )
+        )
+      )(
+        e => IO(e.stop())
+      ).use { _ =>
+        for {
+          aggregates <- mkResources[IO](nbGood, nbBad, localstackPort = Containers.localStackPort).use { r =>
```
pondzix, line 69: Is it possible to have single `use` in this test? So sth like: 

```scala

      val testResources = mkEnrich( 
        configPath = "modules/kinesis/src/it/resources/enrich/count.hocon",
        loggerName = "count"
      )
        .flatMap(_ => mkOtherResources[IO](nbGood, nbBad ....))
      
        testResources.use { 
          //do something
        }

File: modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala [hunk 21]
```diff
@@ -0,0 +1,110 @@
...
+            configPath = "modules/kinesis/src/it/resources/enrich/stop.hocon",
+            loggerName = "stop",
+            network = Some(Containers.localstackNetwork)
+          )
+        )
+      )(
+        e => IO(e.stop()) // in case SIGTERM failed to stop the container
+      ).use { enrich =>
+        for {
+          _ <- IO.sleep(20.seconds) // give Enrich more time to initialize (Kinesis)
```
pondzix, line 95: What about that idea of replacing sleeping with waiting for a specific log? Is it doable?

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+          _ <- IO(println("stop - Sending signal"))
+          _ <- IO(enrich.getDockerClient().killContainerCmd(enrich.getContainerId()).withSignal("TERM").exec())
+          _ <- IO(enrich.isRunning()).retryingOnFailures(
```
pondzix, line 98: Let's extract this some method like `waitUntilStopped`. Retry policy could be also moved there maybe.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
```diff
@@ -0,0 +1,110 @@
...
+
+  override protected val Timeout = 10.minutes
+
+  val retryPolicy =
+    RetryPolicies.limitRetriesByCumulativeDelay(5.minutes, RetryPolicies.capDelay[IO](2.second, RetryPolicies.fullJitter[IO](1.second)))
+
+  def afterAll: Unit = Containers.localstack.stop()
+
+  "enrich-kinesis" should {
+    "initialize with the minimal config" in {
```
pondzix, line 41: Oh, so if it's about parsing config only - can we mention it in the test name?

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: Here are the constraints:
- It's possible to run just one of tests (e.g. `sbt "project kinesis" "IntegrationTest/testOnly com.snowplowanalytics.snowplow.enrich.kinesis.EnrichKinesisSpec testOnly -- ex minimal"`
- If only `minimal` is executed, we don't want to start `localstack`
- If both `count` and `stop` are executed, we want to share `localstack`

My hack for that was to use `lazy val` and to use localstack's network for `count` and `stop`, but I also don't like that, I'll come up with something better.

On top of that because of the `def afterAll: Unit = Containers.localstack.stop()` `localstack` container starts to get created before when the tests exits.

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
pondzix: Can we add a new log? I think it would be nice to have sth like `Enrich initialized properly, ready to process some data`

benjben: Yes we can do that, but at that point Kinesis SDK won't be fully initialized, as it takes some time to initialize KCL with DynamoDB, listing the shards, syncing etc. and I'd like as much of this to have happened when I send the `SIGTERM`, so that they are the more things that could possibly fail to exit, and we catch it in the tests.

benjben: I used `metrics.good` that comes later in the initialization

modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/EnrichKinesisSpec.scala
benjben: Great news is that we can call `stop()` on a non-running container, that made my life easier.

pondzix (approved):
Awesome, great job @benjben!

PR #691, opened 2022-09-12
Retry EOF for blaze client

pondzix:


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -69,6 +70,19 @@ object Clients {
       .withRequestTimeout(readTimeout)
       .withMaxTotalConnections(maxConnections)
       .resource
+      .map(Retry[F](retryPolicy, redactHeadersWhen))
+
+  private def retryPolicy[F[_]] =
+    RetryPolicy[F](
+      RetryPolicy.exponentialBackoff(maxWait = 10.seconds, maxRetry = 10),
```
istreeter, line 77: Do I understand right that we think the first retry will have a very high likelihood of success, even if we retry after no delay?  And am I right in the old implementation of blaze client, it would retry immediately without a delay?

I checked [the http4s implementation of exponentialBackoff](https://github.com/http4s/http4s/blob/v0.21.33/client/src/main/scala/org/http4s/client/middleware/Retry.scala#L175-L179).  It looks like the first retry happens after ~1000 milliseconds (with random factor), and it's not configurable.  That's _probably_ OK if it's a rare exception.  But I wonder if we can do better to avoid adding latency to the output of enrich.  Is it worth considering if we should use a custom retry policy that retries almost immediately?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -69,6 +70,19 @@ object Clients {
...
+      retriable = {
+        case (_, Left(Command.EOF)) => true
+        case _ => false
+      }
+    )
+
+  private def redactHeadersWhen(header: CaseInsensitiveString) =
+    (Headers.SensitiveHeaders + "apikey".ci).contains(header)
```
istreeter, line 85: I assume this is because http4s automatically prints something to the logs in case of a retry.  Please can you share an example of what gets printed?  And which log level is it at?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -58,7 +59,7 @@ object Clients {
       }
     }
 
-  def mkHttp[F[_]: ConcurrentEffect](
+  def mkHttp[F[_]: ConcurrentEffect: Timer]( //TODO it looks like it's used anywhere http client is needed in enrich, so same client everywhere. Maybe we should split clients and make retries only for iglu lookup...?
```
istreeter, line 62: I don't have a problem with letting all clients use this retry policy.  From what I understand, the error happens when a server closes a connection unexpectedly.  It seems that could affect any api call, not just iglu.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
pondzix: I agree, we should try to retry immediately.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
pondzix: [Here](https://github.com/http4s/http4s/blob/382bcb8419cb050c7d38df1d17327da60e7e59b8/client/src/main/scala/org/http4s/client/middleware/Retry.scala#L40) is the code with logging. It's all info log level.

In the case of EOF and retry attempt:

```
[pool-1-thread-1] INFO org.http4s.client.middleware.Retry - Request threw an exception on attempt #1. Retrying after 0 milliseconds
EOF
```
Request is not printed. But looking at the code, when we exhaust attempts it would be printed with headers. It's very unlikely it happens, but still possible! 

When it actually happens, sensitive data is printed as: `apikey: <REDACTED>`

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -69,6 +70,24 @@ object Clients {
...
+      backoff,
+      retriable = {
+        case (_, Left(Command.EOF)) => true
+        case _ => false
+      }
+    )
+
+  //retry immediately, max 10 times
+  private def backoff(attemptNumber: Int): Option[FiniteDuration] =
+    if (attemptNumber > 10) None
```
istreeter, line 86: I paused for a long time on this number "10".  I wonder if it should instead be "1".

Because the specific error case we know about is fixed by a tcp re-connect.  If it is not fixed by the first re-connect, then that's a _new_ error case that we don't know about.  And I wondered if it's better to let such a case raise an error, so at least we discover it.

But it's a tough call, and probably could be argued either way.  I'm happy to approve this as-is.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
```diff
@@ -69,6 +70,24 @@ object Clients {
       .withRequestTimeout(readTimeout)
       .withMaxTotalConnections(maxConnections)
       .resource
+      .map(Retry[F](retryPolicy, redactHeadersWhen))
+
+  private def retryPolicy[F[_]] =
+    RetryPolicy[F](
+      backoff,
+      retriable = {
+        case (_, Left(Command.EOF)) => true
```
istreeter, line 79: Please can make sure there is a github issue that describes this change, and then link to it with a code comment?  This is exactly the kind of code that looks very strange in 5 years time when we've forgotten why we did it.

istreeter (approved):

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/Clients.scala
istreeter: Yeah I think we're OK.  It is used for:

- Telemetry.  Which uses the snowplow scala tracker and so retries requests already, even without this change.
- Metadata (experimental feature).  Same as telemetry, this uses the snowplow scala tracker to relay snowplow events.
- Remote adapter.  These should be idempotent GET requests.
- Downloading assets.  Which also does retries already, even before this change.

PR #693, opened 2022-09-14
Log misconfigured enrichment name

pondzix:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentRegistry.scala
```diff
@@ -90,7 +90,7 @@ object EnrichmentRegistry {
                        sd <- EitherT.fromEither[F](
                                SelfDescribingData.parse(json).leftMap(pe => NonEmptyList.one(pe.code))
                              )
-                       _ <- client.check(sd).leftMap(e => NonEmptyList.one(e.asJson.noSpaces))
+                       _ <- client.check(sd).leftMap(e => NonEmptyList.one(s"Enrichment with key '${sd.schema.toSchemaUri}` is invalid - ${e.asJson.noSpaces}"))
```
pondzix, line 93: Example log: 

```
[pool-1-thread-2] ERROR com.snowplowanalytics.snowplow.enrich.common.fs2.Run - CLI arguments valid but some of the configuration is not correct. Error: Cannot decode enrichments - Enrichment with key 'iglu:com.snowplowanalytics.snowplow/ua_parser_config/jsonschema/1-0-0` is invalid - {"error":"ValidationError","dataReports":[{"message":"$.parameters: may only have a maximum of 0 properties","path":"$.parameters","keyword":"maxProperties","targets":["0"]}]}
```

This is useful when `data` content of enrichment is not valid according to specific enrichment schema. But... processing enrichement config can fail before calling `check` on client, e.g. when provided json is invalid SDJ (`SelfDescribingData.parse` operation). In that case, current solution wouldn't provide enrichment name in logs. 

Maybe we should carry file name when loading enrichment jsons from directory (or item id when loading from dynamodb) and use it in logs instead of using schema key extracted from json? Then we could probably cover every failure.

istreeter: How much extra work is it to do what you suggest here?

Honestly, I think the change you've made already is excellent and covers the situation we had problems with before.  Debugging invalid data is more difficult debugging an invalid SDJ.  So I'm fine for you to merge it as is.

On the other hand, if there is an opportunity to make a quick easy code change to cover the other case you mention -- then we should definitely take that opportunity whilst we are thinking about it.  You just need to assess how quick and how easy it would be.

pondzix: I forgot about the case when we load enrichments from base64 json. Then we don't have any way to name an invalid json other than schema key, right? I'm going to leave it as is.

PR #696, opened 2022-09-29
Common: remove url decoding the user agent

spenes:


istreeter (commented):
Please can you just confirm, what would happen if I put a percent-encoded user agent in the query string like this:

```
curl http://<COLLECTOR>t/i?e=pp&aid=test&p=web&tna=test&ua=foo%3B%20bar%3B%20baz
```

Does it get treated exactly the same as it did before this change?

spenes:
@istreeter Yep, I can confirm that. Nothing changed on that behavior.

istreeter (approved):

PR #708, opened 2022-10-18
Bump specs2

benjben:


project/Dependencies.scala
```diff
@@ -252,7 +255,8 @@ object Dependencies {
       specs2Cats,
       specs2Scalacheck,
       specs2Mock,
-      circeLiteral % Test
+      circeLiteral % Test,
+      parserCombinators
```
istreeter, line 259: Should this also be `% Test` ?

istreeter: Oh ignore, I see you did it above.

istreeter (approved):

PR #723, opened 2022-11-01
Release 3.5.1

istreeter:


project/Dependencies.scala
```diff
@@ -196,15 +196,19 @@ object Dependencies {
     // FS2
     val decline          = "com.monovore"                     %% "decline"                               % V.decline
     val fs2PubSub        = "com.permutive"                    %% "fs2-google-pubsub-grpc"                % V.fs2PubSub
-    val fs2Aws           = ("io.laserdisc"                    %% "fs2-aws"                               % V.fs2Aws).exclude("com.amazonaws", "amazon-kinesis-producer")
+    val fs2Aws           = ("io.laserdisc"                    %% "fs2-aws"                               % V.fs2Aws)
+                           .exclude("com.amazonaws", "amazon-kinesis-producer")
+                           .exclude("software.amazon.kinesis", "amazon-kinesis-client")
```
benjben, line 201: Don't the versions of this `amazon-kinesis-client` always get overridden by the ones of `kinesisClient2` ?

istreeter: Yes it gets overridden, but this is to enforce that the `software.amazon.glue` packages are properly excluded [further down](https://github.com/snowplow/enrich/blob/3.5.1-rc2/project/Dependencies.scala#L210-L211).  Without this line, I still end up with `software.amazon.glue` in the final assembly.

benjben: Oh I see !

project/Dependencies.scala
```diff
@@ -43,7 +43,7 @@ object Dependencies {
     val iabClient        = "0.2.0"
     val yauaa            = "5.23"
     val guava            = "28.1-jre"
-    val slf4j            = "1.7.32"
+    val slf4j            = "2.0.3"
```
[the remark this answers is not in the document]
istreeter: I've tested the logging, and the integration test passes.  It also runs in the github action, no?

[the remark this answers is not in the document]
benjben: Ah yes true, if the logging breaks, the integration tests won't pass !

PR #726, opened 2022-11-09
common: remove duplicates from Sendgrid adapter payload

istreeter:


dilyand (approved):
The SendGrid docs [state](https://docs.sendgrid.com/for-developers/tracking-events/troubleshooting#duplicate-events) that:

> It is possible to see duplicate events in the data posted by the Event Webhook.
We recommend that you use some form of deduplication when processing or storing your Event Webhook data using the `sg_event_id` as a differentiator since this ID is unique for every event where `sg_event_id` is present.

AFAICT, `sg_event_id` is sent with every event, so the deduplication with `.toSet` is fine. 

Because of the formulation "this ID is unique for every event where `sg_event_id` is present", I was thinking about cases where it might be missing. They would have also been deduplicated by this method, but might have turned out to be unique events that are just missing this property. But all sources seem to suggest that all events are expected to have it.

PR #727, opened 2022-11-09
Amendment to sendgrid adaptor: deduplicate by sg_event_id

istreeter:


istreeter:
When we merge to **master** (the release), this feature should end up as a single commit and a single item in the changelog.  That's the most important thing, and beyond that I don't have strong opinions about the develop branch.

I figure I might as well force push to the develop branch.  I don't think it will upset anyone.  Anything else will be needlessly following some rules but for no clear benefit.

Shout though if you don't want me to force push to develop.

PR #731, opened 2022-11-23
Parse Iglu resolver as a HOCON instead of JSON

istreeter:


benjben (commented):
That's great improvement already !

There are 3 things configuring Enrich :
- The config file itself, which is read as a HOCON and then parsed as a JSON
- The resolver, that is a JSON now read as a HOCON first and then parsed as JSON
- The enrichments files, [which are read directly as JSONs](https://github.com/snowplow/enrich/blob/master/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ParsedConfigs.scala#L74)

What do you think of using the same mechanism for the enrichments files, i.e. reading then as HOCON first to resolve the environment variables and then parsing them as JSONs ?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala
```diff
@@ -86,19 +83,16 @@ object ConfigFile {
...
     deriveConfiguredEncoder[ConfigFile]
 
-  def parse[F[_]: Sync: ContextShift](in: EncodedHoconOrPath): EitherT[F, String, ConfigFile] = {
-    val cliSource = in.fold(_.value, ConfigSource.file(_))
-    val source = ConfigSource.default(cliSource.withFallback(ConfigSource.default))
+  /* Defines where to look for default values if they are not in the provided file
+   *
+   * Command line parameters have the highest priority.
+   * Then the provided file.
+   * Then the application.conf file.
```
benjben, line 90: :+1:

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ParsedConfigs.scala
```diff
@@ -97,6 +99,17 @@ object ParsedConfigs {
       _ <- EitherT.liftF(Logger[F].info(show"Parsed following enrichments: ${configs.map(_.schemaKey.name).mkString(", ")}"))
     } yield ParsedConfigs(igluJson, configs, configFile, goodPartitionKey, piiPartitionKey, goodAttributes, piiAttributes)
 
+  private[config] def parseEncodedOrPath[F[_]: Sync, A: Decoder](
+    in: EncodedHoconOrPath,
+    fallbacks: TSConfig => TSConfig
+  ): EitherT[F, String, A] =
+    in match {
+      case Left(v) =>
```
benjben, line 107:
```suggestion
      case Left(hocon) =>
```

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
```diff
@@ -40,18 +43,29 @@ object FileSystem {
...
 
-  def readJson[F[_]: Sync](path: Path): EitherT[F, String, Json] =
+  def readJson[F[_]: Sync, A: Decoder](path: Path, fallbacks: TSConfig => TSConfig): EitherT[F, String, A] =
     Sync[F]
-      .delay[String](Files.readString(path))
+      .delay(SSource.fromFile(path.toFile).mkString)
       .attemptT
-      .leftMap(e => show"Error reading ${path.toAbsolutePath.toString} JSON file from filesystem: ${e.getMessage}")
-      .subflatMap(str => parse(str).leftMap(e => show"Cannot parse JSON in ${path.toAbsolutePath.toString}: ${e.getMessage()}"))
+      .leftMap(e => s"Cannot read file: ${e.getMessage}")
```
benjben, line 50:
```suggestion
      .leftMap(e => s"Cannot read file $path: ${e.getMessage}")
```

modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/DynamoDbConfig.scala
```diff
@@ -67,9 +66,9 @@ object DynamoDbConfig {
 
   private def updateArg[F[_]: ContextShift: Sync: Timer](
     blocker: Blocker,
-    orig: EncodedOrPath,
-    getConfig: (Blocker, String, String, String) => F[Base64Json]
-  ): F[EncodedOrPath] =
+    orig: Either[Base64Hocon, Path],
```
benjben, line 69:
```suggestion
    orig: EncodedHoconOrPath,
```
Is there any particular reason why not using it?

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/Base64HoconSpec.scala
```diff
@@ -26,10 +26,10 @@ class Base64HoconSpec extends Specification {
...
-    "parse a plain string as a HOCON source" in {
-      val inputStr = "+"
+    "not parse any other base-64 encoded string as a HOCON source" in {
+      val inputStr = "xyz"
       val input = getEncoder.encodeToString(inputStr.getBytes())
-      Argument[Base64Hocon].read(input).toEither must beRight
+      Argument[Base64Hocon].read(input).toEither must beLeft
     }
   }
 }
```
benjben, line 35: Neat: add a test with a hocon file (in `test/resources`) that is to be resolved and check that it is [when the environment variable is set](https://stackoverflow.com/questions/37330416/set-default-env-variable-for-test-configuration-in-sbt).

benjben (approved):
It looks great @istreeter !

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
istreeter: This was actually deliberate.  I found that `e.getMessage` already contains the path name.  And as this is a very user-facing error message (e.g. when a developer is trying to get enrich working for the first time) I thought it was important to have a plain error message that is easy to read.  So in my opinion it's more helpful to omit the extra `$path` from the message.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/Base64Hocon.scala
```diff
@@ -36,8 +39,17 @@ object Base64Hocon {
     }
 
   def parseHocon(str: String): Either[String, Base64Hocon] =
-    Either
-      .catchOnly[IllegalArgumentException](base64.decode(str))
-      .map(bytes => Base64Hocon(ConfigSource.string(new String(bytes))))
-      .leftMap(_.getMessage)
+    for {
+      bytes <- Either.catchOnly[IllegalArgumentException](base64.decode(str)).leftMap(_.getMessage)
+      tsConfig <- Either.catchNonFatal(ConfigFactory.parseString(new String(bytes))).leftMap(_.getMessage)
```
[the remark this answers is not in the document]
istreeter: Yes we should!!!  Can't believe I fell in that mistake.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
benjben: Agreed! As long as there is the path we're good!

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/Base64HoconSpec.scala
istreeter: Added 2 tests.

PR #737, opened 2023-01-02
Add UA parser enrichment cache

pondzix:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/UaParserEnrichment.scala
```diff
@@ -108,7 +106,22 @@ final case class UaParserEnrichment(schemaKey: SchemaKey, parser: Parser) extend
...
+          .pure[F]
+          .flatTap {
+            case Right(context) =>
+              userAgentCache.put(useragent, context)
+            case Left(_) =>
+              Monad[F].unit
+          }
+    }
+
+  private def evaluateUserAgentContext(useragent: String) =
```
istreeter, line 124: Stylistic thing: I have a preference for putting the return type, even when the method is private.

```
private def evaluateUserAgentContext(useragent: String): Either[FailureDetails.EnrichmentFailure, SelfDescribingData[Json]] =
```

istreeter (approved):
Looks good, well done.

PR #739, opened 2023-01-18
Enrich Nsq

spenes:


istreeter (commented):
Very nice!

Sorry I'm very late to suggest this.... I wonder if there is any benefit of switching to [an alternative nsq client lib](https://github.com/sproutsocial/nsq-j)?  Currently we are using [our own fork of a nsq client](https://github.com/snowplow/JavaNSQClient), but it might be nice if we had one less thing to maintain (even though we don't do much to maintain it).

spenes:
Thanks @istreeter!

I guess it wouldn't be too hard to switch to that library but I am not sure if that would benefit much. As you said, we don't do much to maintain our nsq client fork. Also, I am not sure if we will continue to support enrich-nsq for a long time since we are using it only on the Snowplow Mini and Snowplow Mini's future still isn't certain. Therefore, I would vote to not replace the library to not create any additional work in here. But, I am not strongly opinionated about it. If you insist it is better to replace it, I am okay with that.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+        lookup.addLookupAddress(config.lookupHost, config.lookupPort)
+        val consumer = new NSQConsumer(
+          lookup,
+          config.topic,
+          config.channel,
+          messageCallback,
+          new NSQConfig(),
+          errorCallback
+        )
+        consumer.start()
```
benjben, line 83: It seems that only this line is effectful and should be in `Resource.make`

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+        val lookup = new DefaultNSQLookup
+        lookup.addLookupAddress(config.lookupHost, config.lookupPort)
+        val consumer = new NSQConsumer(
+          lookup,
+          config.topic,
+          config.channel,
+          messageCallback,
+          new NSQConfig(),
+          errorCallback
+        )
```
benjben, line 82: Can we put that in `createConsumer` function ?

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+        consumer.start()
+      }
+    )(service =>
```
benjben, line 85:
```suggestion
    )(consumer =>
```

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+      blocker
+        .delay(service.shutdown())
+        .handleErrorWith(e => Logger[F].error(s"Cannot terminate NSQ consumer ${e.getMessage}"))
```
benjben, line 88:
```suggestion
        .handleErrorWith(e => Logger[F].error(e)(s"Cannot terminate NSQ consumer"))
```

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+    queue: Queue[F, Either[Throwable, Record[F]]],
+    config: Input.Nsq
+  ): Resource[F, NSQConsumer] =
+    Resource.make(
+      Sync[F].delay {
+        val messageCallback = new NSQMessageCallback {
+          override def message(message: NSQMessage): Unit = {
+            val msgBytes = message.getMessage
+            val enqueue = queue.enqueue1(Right(Record(msgBytes, Sync[F].delay(message.finished()))))
+            ConcurrentEffect[F].toIO(enqueue).unsafeRunSync()
```
benjben, line 64: Are you using `ConcurrentEffect` so that the enqueing can be cancelled or only to get an `IO` that we can run? In the latter case, I think that we can just use `IO` everywhere instead of generic `F`.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Sink.scala
```diff
@@ -0,0 +1,77 @@
...
+  private def sinkBatch[F[_]: Sync: ContextShift](
+    blocker: Blocker,
+    producer: NSQProducer,
+    topic: String
+  )(
+    records: List[AttributedData[Array[Byte]]]
+  ): F[Unit] =
+    blocker.delay {
+      producer.produceMulti(topic, records.map(_.data).asJava)
+    }
```
benjben, line 76: Would be nice to add some retrying mechanism

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Sink.scala
```diff
@@ -0,0 +1,77 @@
...
+      case c =>
+        Resource.eval(Sync[F].raiseError(new IllegalArgumentException(s"Output $c is not NSQ")))
+    }
+
+  private def createNsqProducer[F[_]: Sync: ContextShift](blocker: Blocker, config: Output.Nsq): Resource[F, NSQProducer] =
+    Resource.make(
+      Sync[F].delay {
+        val producer = new NSQProducer()
+        producer.addAddress(config.nsqdHost, config.nsqdPort)
+        producer.start()
```
benjben, line 59: Same as for the consumer, only `producer.start()` should be in `Resource.make`

benjben (commented):
It looks great already @spenes !

config/config.nsq.minimal.hocon
```diff
@@ -0,0 +1,22 @@
+{
+  "input": {
+    "topic": "collector-payloads"
+    "channel": "collector-payloads-channel"
+    "lookupHost": "127.0.0.1"
+    "lookupPort": 4161
```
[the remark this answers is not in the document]
spenes: I think it would be better to specify what is the host address for in here. If we make it `nsqdHost`, it wouldn't be correct for input part.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
```diff
@@ -0,0 +1,90 @@
...
+  def init[F[_]: ConcurrentEffect: ContextShift](
+    blocker: Blocker,
+    input: Input
+  ): Stream[F, Record[F]] =
+    input match {
+      case config: Input.Nsq =>
+        for {
+          q <- Stream.eval(Queue.bounded[F, Either[Throwable, Record[F]]](config.maxBufferQueueSize))
+          _ <- Stream.resource(startConsumer(blocker, q, config))
+          msg <- q.dequeueChunk(config.maxBufferQueueSize).flatMap(m => Stream.fromEither[F](m))
```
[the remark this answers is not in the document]
spenes: Yep, exactly.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
spenes: If I am not missing something, there is no 'createConsumer' function at the moment in that file. May I ask what you meant exactly ?

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
spenes: It is for getting an `IO` to run it. 

I see your point but using `F` still looks better to me. If you aren't strongly opinionated about it, I would vote to leave it as it is.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
spenes: May I ask what is the reason to do that exactly ? To my understanding, all the object initializations are also the part of resource creation process therefore it is better to do them in the `Resource.make`.  Does it have advantage to do them outside of `Resource.make` ?

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Sink.scala
spenes: Same question: https://github.com/snowplow/enrich/pull/739#discussion_r1087294865

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
istreeter: For what it's worth, [this is the equivalent line in fs2-aws](https://github.com/laserdisc-io/fs2-aws/blob/v3.1.1/fs2-aws/src/main/scala/fs2/aws/kinesis/Kinesis.scala#L86).

fs2-google-pubsub does not have an equivalent line, because it uses a `java.util.concurrent.BlockingQueue` instead of a fs2 queue ([here](https://github.com/permutive-engineering/fs2-google-pubsub/blob/v0.18.1/fs2-google-pubsub-grpc/src/main/scala/com/permutive/pubsub/consumer/grpc/internal/PubsubSubscriber.scala#L68)).

When we bump to CE3 then this line will look slightly different.  You would use a `Dispatcher[F]` ([docs](https://typelevel.org/cats-effect/docs/std/dispatcher)) and then:

```
dispatcher.unsafeRunSync(enqueue)
```

Anyway, I don't have a point to make.  I just thought it was interesting.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
benjben: It's just to help the understanding when reading of the code.

> To my understanding, all the object initializations are also the part of resource creation process

The way I see it, only what is effectful should be in `Resource.make`.

It would help someone not familiar with the code to see quickly that only `consumer.start()` is effectful.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
benjben: Yes exactly, what I'm saying is that it would be great to create this function and put this inside.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
benjben: I'm not strongly opinionated and I'm fine with leaving this as it is. There is no great solution anyway to mix functional and non-functional code.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
spenes: I see your point but I think it is hard to judge what is effectful and what is not here. It is possible there is some state in the initialized objects and if we create them outside of Resource, it is possible they will be reused by different Resources and this might lead to subtle errors. We would be eliminate this if we create everything in the `Resource.make`.

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala
spenes: Ah okay, will do that 👍

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Containers.scala
```diff
@@ -0,0 +1,274 @@
...
+      n => Sync[F].delay(n.close())
+    )
+
+  private def nsqlookupd[F[_]: Sync](
+    network: Network,
+    networkInfo: NetworkInfo
+  ): Resource[F, JGenericContainer[_]] =
+    Resource.make (
+      Sync[F].delay {
+        val container = FixedHostPortGenericContainer(
```
benjben, line 137: We shouldn't need to use fixed ports, we can use `container.getMappedPort`, like [here](https://github.com/snowplow/enrich/blob/master/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Containers.scala#L70)

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Utils.scala
```diff
@@ -0,0 +1,135 @@
...
+    for {
+      e <- Sync[F].delay(Event.parse(new String(r.data)).getOrElse(throw new RuntimeException("can't parse enriched event")))
+      _ <- r.ack
+      _ <- ref.update(updateAggregateGood(_, e))
+    } yield ()
+  }
+
+  def aggregateBad[F[_]: Sync](r: Record[F], ref: Ref[F, AggregateBad]): F[Unit] = {
+    for {
+      br <- Sync[F].delay(new String(r.data))
```
benjben, line 124: Nit: parse as `BadRow`. Example [here](https://github.com/snowplow/enrich/blob/master/modules/kinesis/src/it/scala/com/snowplowanalytics/snowplow/enrich/kinesis/utils.scala#L94-L99) (should be moved to `common-fs2` if we reuse it here)

benjben (commented):
Looks great @spenes !

I have mostly one question about testing topolgy.

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Utils.scala
```diff
@@ -0,0 +1,135 @@
...
+  val backoffPolicy = BackoffPolicy(
+    minBackoff = 100.milliseconds,
+    maxBackoff = 10.seconds,
+    maxRetries = Some(10)
+  )
+
+  type AggregateGood = List[Event]
+  type AggregateBad = List[String]
+
+  def mkResources[F[_]: Async: ContextShift: Timer] =
```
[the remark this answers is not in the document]
spenes: I guess no harm to use it like that. I would vote to leave it as it is.

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Containers.scala
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * Integration test tries to access nsqd from outside of docker network
+   * therefore it can't access it with docker service name.
+   * Since nsqd broadcasts its address through nsqlookup, it is impossible to
+   * broadcast its address as localhost and docker service name at the same time
+   * therefore we've created second nsqd and nsqlookup instances that will be used by
+   * integration tests and NSQ messages sent to first nsqd instance will be replicated
+   * to second nsqd instance with nsq_to_nsq tool.
+   */
```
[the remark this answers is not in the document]
spenes: The problem is lookup operation in NSQ. NSQ consumer connects to [nsqlookupd](https://github.com/snowplow/enrich/blob/enrich-nsq/modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/Source.scala#L89) to discover nsqd instances. And [nsqd](https://nsq.io/components/nsqd.html) broadcasts its address through nsqlookup. Then NSQ consumer tries to connect to the nsqd client through brodcasted address. The problem in here is that NSQ applications and Enrich-NSQ is running on same Docker network. However, tests are running outside of that network. Therefore, if we make the nsqd broadcast address `localhost`, Enrich NSQ consumer in the Docker container can't access to it because they aren't sharing same localhost. It can only access to nsqd via Docker network alias. In order to solve this problem, we can make the nsqd broadcast address Docker network alias. In that case however, integration tests can't access to nsqd because they are not in Docker network.

In order to solve this problem, I've created two nsqd instances. The broadcast address of the one of them is `localhost`. And other's broadcast address is its Docker network alias. Messages are relayed one to another with nsq_to_nsq. With this way, Enrich-NSQ in Docker container can use the nsqd with Docker network alias. Then, its content will be relayed to nsqd instance with localhost as broadcast address. Tests will consume messages from this nsqd instance.

I know this is quite complicated and solution isn't straighforward. But this was the only solution I can find. Please let me know if you have something better in mind. Also, we can have 1-1 if you have still some questions.

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Containers.scala
spenes: We have a problem similar to the one explained [here](https://github.com/snowplow/enrich/pull/739#discussion_r1095774056). [nsqd](https://nsq.io/components/nsqd.html) needs to broadcast its port as well. If we don't use fixed ports, it isn't possible to know the assigned port before container started.

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Containers.scala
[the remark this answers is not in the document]
benjben: Thanks for the detailed explanations, I get it now! I can't think of an easier alternative.

modules/common-fs2/src/it/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/test/Utils.scala
```diff
@@ -0,0 +1,31 @@
+/*
+ * Copyright (c) 2022-2022 Snowplow Analytics Ltd. All rights reserved.
```
benjben, line 2: Time goes fast :)

modules/nsq/src/it/resources/enrich/enrich-nsq.hocon
```diff
@@ -0,0 +1,52 @@
...
+      "cloudwatch": false
+    }
+  }
+
+  "telemetry": {
+    "disable": true
+  }
+
+  "featureFlags": {
+    "acceptInvalid": true
```
benjben, line 50: Is there a reason why we want to accept invalid events ?

modules/nsq/src/it/resources/enrich/enrich-nsq.hocon
```diff
@@ -0,0 +1,52 @@
...
+      # Optional. Log to stdout using Slf4j
+      "stdout": {
+        "period": "10 seconds"
+
+        # Optional, override the default metric prefix
+        # "prefix": "snowplow.enrich."
+      }
+
+      # Optional. Send KCL and KPL metrics to Cloudwatch
+      "cloudwatch": false
```
benjben, line 41:
```suggestion
```

It's already [in the defaults](https://github.com/snowplow/enrich/blob/enrich-nsq/modules/nsq/src/main/resources/application.conf#L44-L48).

modules/nsq/src/it/resources/enrich/enrich-nsq.hocon
```diff
@@ -0,0 +1,52 @@
...
+      "nsqdHost": ${NSQD_HOST}
+      "nsqdPort": ${NSQD_PORT}
+    }
+  }
+
+  "monitoring": {
+
+    "sentry": {
+      "dsn": ""
+    }
```
benjben, line 27: It's [optional](https://github.com/snowplow/enrich/blob/master/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala#L399) so you can omit this block

modules/nsq/src/it/scala/com/snowplowanalytics/snowplow/enrich/nsq/test/Containers.scala
```diff
@@ -218,21 +218,24 @@ object Containers {
...
-          nsqdHost = topology.nsqd1.networkAlias,
-          nsqdPort = topology.nsqd1.tcpPort,
-          inputTopic = topology.sourceTopic,
-          goodOutputTopic = topology.goodDestTopic,
-          badOutputTopic = topology.badDestTopic
-        )
         val container = GenericContainer(
           dockerImage = s"snowplow/snowplow-enrich-nsq:${BuildInfo.version}-distroless",
           env = Map(
             "JAVA_OPTS" -> "-Dorg.slf4j.simpleLogger.defaultLogLevel=info",
```
benjben, line 224: I know that this is what we use for Kinesis tests, but I think that now that we use the `-distroless` images for the tests this is not correct any more and we have to use `JDK_JAVA_OPTIONS` instead. Could you check and update Kinesis as well if that's the case please?

modules/nsq/src/it/resources/enrich/enrich-nsq.hocon
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  "telemetry": {
+    "disable": true
+  }
+
+  "featureFlags": {
+    "acceptInvalid": false
+  }
```
benjben, line 43:
```suggestion
```
Already in the default.

benjben (approved):
Looks great @spenes !

PR #748, opened 2023-02-16
common: improve caching in API/SQL enrichments (close #747)

### LINKED ISSUE #747: common: improve caching in  API/SQL enrichments
If we get an error from external API/database (e.g. because it's down), such error is cached by enrich. It's cached because we don't want to overload API/database experiencing some issues. 

TTL for errors stored in the cache is the same as for successful results. If this TTL is some big integer (which is often reasonable in case of successful/valid results) we can get stuck with errors in the cache for while. So even though external API, which originally produced errors,  is now working correctly we don't know that cause we use errors stored in the cache. 

Potential improvements: 

* Use different, smaller TTL for errors (make it configurable?).
* For scenario like: if successful result is stored in cache => TTL expires => new value is fetched => we get an error. In such case store error in the cache (with smaller TTL), but return last known 'old' good value to further processing. This fallback would allow enrich to produce fewer badrows in case of 'getting stuck' with errors in the enrichment cache. 

Those improvements can be implemented for both API and SQL enrichments.

It all resembles behaviour of circuit-breaker. Adding proper circuit-breaker implementation (most likely coming from some external lib) could be problematic as we still have to support `Id` monad, hence improvements for existing solution relying on errors caching.

pondzix:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
```diff
@@ -212,43 +202,28 @@ sealed trait CreateApiRequestEnrichment[F[_]] {
...
     new CreateApiRequestEnrichment[F] {
-      def create(conf: ApiRequestConf): F[ApiRequestEnrichment[F]] =
-        CLM
-          .create(conf.cache.size)
-          .map(c =>
+      def create(conf: ApiRequestConf): F[ApiRequestEnrichment[F]] = {
+        val cacheConfig = CachingEvaluator.Config(
+          size = conf.cache.size,
+          successTtl = conf.cache.ttl,
+          errorTtl = conf.cache.ttl / 10
```
pondzix, line 213: Wondering if we should make error ttl configurable next to the 'success' ttl or just do it like this, so divide 'success' by some magic number?

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/CachingEvaluator.scala
```diff
@@ -0,0 +1,110 @@
...
+import cats.implicits._
+import com.snowplowanalytics.lrumap.{CreateLruMap, LruMap}
+import com.snowplowanalytics.snowplow.enrich.common.enrichments.registry.CachingEvaluator.{Cache, CachedItem, GetResult, Value}
+
+import java.util.concurrent.TimeUnit
+
+final class CachingEvaluator[F[_]: Clock: Monad, K, V](
+  cache: Cache[F, K, V],
+  config: CachingEvaluator.Config
+) {
```
istreeter, line 14: The pattern we use elsewhere in Snowplow code is that classes do not have implicit type classes like `Clock` or `Monad`.  And instead we put the type class on the methods:

```
final class CachingEvaluator[F[_], K, V](....) {
  def evaluateForKey(....)(implicit M: Monad[F], C: Clock[F]): F[Either[Throwable, V]] =
    ???
}
```

To be completely honest, the people who cared most about this pattern have now left Snowplow!!  So in a way we are free to ignore that rule and do what we want :)  And unfortunately I don't know the full details of why they liked that pattern.

Would it be problematic to use the old pattern here too?

pondzix: It wouldn't be problematic at all. If it's the pattern, let's be consistent! I will change it.

spenes (approved):
LGTM!

PR #750, opened 2023-02-28
Schema Override

spenes:
Details in https://github.com/snowplow/enrich/issues/751


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
```diff
@@ -259,20 +258,48 @@ object IgluUtils {
   private def parseAndValidateSDJ_sv[F[_]: Monad: RegistryLookup: Clock]( // _sv for SchemaViolation
```
voropaevp, line 258: mixed naming convention. `parseAndValidateSdjSv`

voropaevp, line 258: mixed naming convention. `parseAndValidateSdjSv`

voropaevp (approved):
LGTM

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/IgluUtils.scala
voropaevp, line 258: mixed naming convention

PR #756, opened 2023-03-16
common-fs2: fix env var substitution for JSON files (close #753)

### LINKED ISSUE #753: common-fs2: fix env var substitution for JSON files
**Project**:  Stream Enrich

**Version**: 3.7.0

**Expected behavior**:

Since version 3.7.0 using environment variables substitutions should not trigger an error seen below. The config should be parsed.

**Actual behavior**:

```
[pool-1-thread-1] ERROR com.snowplowanalytics.snowplow.enrich.common.fs2.Run - CLI arguments valid but some of the configuration is not correct. Error: Cannot parse file /home/snowplow/resolver/resolver.conf: ParsingFailure: need to Config#resolve(), see the API docs for Config#resolve(); substitution not resolved: ConfigConcatenation("https://"${domain}"/iglu-foo/")
```

It happens with `resolver.hocon` as well as all the other enrichments configs.

**Steps to reproduce**:

 1. Prepare directory with configs `/test`
`/test/resolver/resolver.json`
```
{
  "schema": "iglu:com.snowplowanalytics.iglu/resolver-config/jsonschema/1-0-1",
  "data": {
    "cacheSize": 500,
    "repositories": [
      {
        "name": "foo",
        "priority": 0,
        "vendorPrefixes": ["com.foo"],
        "connection": {
          "http": {
            "uri": "https://"${domain}"/iglu-foo/"
          }
        }
      }
    ]
  }
}

```

`/test/config.hocon`

```
{
  "input": {
    "streamName": "collector-payloads"
  }

  "output": {
    "good": {
      "streamName": "enriched"
    }

    "bad": {
      "streamName": "bad"
    }
  }
}

```

`/test/enrichments/` just empty dir
 2. Run enricher with

```
docker run \
  -it --rm \
  -v /test:/snowplow \
  -e domain="foo" \
  snowplow/snowplow-enrich-kinesis:3.7.0 \
  --enrichments /snowplow/enrichments \
  --iglu-config /snowplow/resolver/resolver.hocon \
  --config /snowplow/config.hocon


```

benjben:
See #753

### DIFF
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
@@ -51,6 +51,7 @@ object FileSystem {
       .subflatMap { text =>
         val either = for {
           tsConfig <- Either.catchNonFatal(ConfigFactory.parseString(text)).leftMap(_.getMessage)
+          tsConfig <- Either.catchNonFatal(tsConfig.resolve()).leftMap(e => s"Can't resolve config: ${e.getMessage}")
           tsConfig <- Either.catchNonFatal(fallbacks(tsConfig)).leftMap(_.getMessage)
           parsed <- tsConfig.as[A].leftMap(_.show)
         } yield parsed


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/io/FileSystem.scala
```diff
@@ -51,8 +51,9 @@ object FileSystem {
       .subflatMap { text =>
         val either = for {
           tsConfig <- Either.catchNonFatal(ConfigFactory.parseString(text)).leftMap(_.getMessage)
-          tsConfig <- Either.catchNonFatal(fallbacks(tsConfig)).leftMap(_.getMessage)
-          parsed <- tsConfig.as[A].leftMap(_.show)
+          resolved <- Either.catchNonFatal(tsConfig.resolve()).leftMap(e => s"Can't resolve config: ${e.getMessage}")
+          withFallbacks <- Either.catchNonFatal(fallbacks(resolved)).leftMap(_.getMessage)
+          parsed <- withFallbacks.as[A].leftMap(_.show)
```
istreeter, line 56: This might be controversial, but if it was me I would write it like this, shadowing the `tsConfig` parameter.

```
    tsConfig <- Either.catchNonFatal(ConfigFactory.parseString(text)).leftMap(_.getMessage)
    tsConfig <- Either.catchNonFatal(tsConfig.resolve()).leftMap(e => s"Can't resolve config: ${e.getMessage}")
    tsConfig <- Either.catchNonFatal(fallbacks(tsConfig)).leftMap(_.getMessage)
    parsed <- tsConfig.as[A].leftMap(_.show)
```

I like how on each line there is only one thing in scope that has type `Config`.

And as someone reading the code I find it easier to quickly see that each `tsConfig` is used once and then never used again.  As we add more names for things, it takes my brain longer to work out what the code does.

istreeter (approved):

PR #758, opened 2023-03-20
Ignore API/SQL enrichments on error

pondzix:
Changes in the iglu central schemas - https://github.com/snowplow/iglu-central/pull/1273

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
@@ -13,7 +13,7 @@
 package com.snowplowanalytics.snowplow.enrich.common.enrichments.registry.apirequest
 
 import cats.Monad
-import cats.data.{EitherT, NonEmptyList, ValidatedNel}
+import cats.data.{EitherT, NonEmptyList, Validated, ValidatedNel}
 import cats.effect.Clock
 import cats.implicits._
 import com.snowplowanalytics.iglu.core.circe.implicits._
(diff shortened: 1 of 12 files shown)


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
```diff
@@ -69,9 +69,16 @@ object ApiRequestEnrichment extends ParseableEnrichment {
           inputs,
           CirceUtils.extract[HttpApi](c, "parameters", "api", "http").toValidatedNel,
           CirceUtils.extract[List[Output]](c, "parameters", "outputs").toValidatedNel,
-          CirceUtils.extract[Cache](c, "parameters", "cache").toValidatedNel
-        ).mapN { (inputs, api, outputs, cache) =>
-          ApiRequestConf(schemaKey, inputs, api, outputs, cache)
+          CirceUtils.extract[Cache](c, "parameters", "cache").toValidatedNel,
+          CirceUtils
+            .extract[Option[Boolean]](c, "parameters", "ignoreOnError")
```
pondzix, line 74: I guess we would have to also include this new param in a new version (1.0.2?) on [iglucentral](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-1)?

istreeter (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
benjben: [Indeed](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/api_request_enrichment_config/jsonschema/1-0-1#L160)

benjben:
Is a timeout also considered as an error ? I think that it should

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
benjben: Have you thought of the roll out / roll back plan ? I wonder if Enrich should be able to parse both versions of the config file until the transition is complete ?

benjben (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/apirequest/ApiRequestEnrichment.scala
pondzix: `supportedSchema` here is: 

```
override val supportedSchema =
    SchemaCriterion(
      "com.snowplowanalytics.snowplow.enrichments",
      "api_request_enrichment_config",
      "jsonschema",
      1,
      0
    )
```

1.0.0, 1.0.1 and 1.0.2 will work. And we parse `ignoreOnError` as `Option[Boolean]` so it's optional with default value `false`.

PR #762, opened 2023-04-03
common: do not validate enrichment names

stanch:
Users are often confused by the “vendor” and “name” fields in the enrichment configs. There is no reason we need to validate or even use these fields, because the “schema” field already contains enough information. This commit removes the validation on “name” (I believe “vendor” is not validated) and only uses “schema” instead.


stanch:
This is best reviewed with whitespace changes hidden

benjben (commented):
Looks good!

What we usually do is to create an issue with the details, and to name the commit with the same name as the issue with `(close #xx)` at the end. See [this commit](https://github.com/snowplow/stream-collector/commit/ea05958a81aa990cca19c601149d59b9772514e7) for [this issue](https://github.com/snowplow/stream-collector/issues/276).

This way of doing is debatable, but I personally like it because in the release notes on Github we always put the changelog (e.g. [here](https://github.com/snowplow/enrich/releases/tag/3.7.0)) and by clicking on an issue we can directly get the context.

stanch:
Done

benjben (approved):
LGTM!

I updated the issue name to have a 1:1 relationship between issue name and commit

stanch:
Oops!

PR #766, opened 2023-04-08
Fix/sql threadpools

istreeter:


benjben:
Sorry for merging @istreeter I had 2 tabs open and didn't merge the right one :man_facepalming: 

Is it fine though ?

istreeter:
Ha!  It's fine.  I mean, this is completely untested -- I only did the fun bit of writing the code.  Ideally we should recreate the problem that we _think_ exists (it's unproven so far), and then show that this new code fixes it.  In any case we need to do that before the release.  But it's fine to merge this into develop for now, as long as you were happy with the shape of the code.

PR #770, opened 2023-04-12
common: Provide derived contexts to JS enrichment (close #769)

### LINKED ISSUE #769: common: Provide derived contexts to the JavaScript enrichment
Currently the JavaScript enrichment allows users to call `event.getDerived_contexts()`, however, it’s always `null`. It would be useful for some enrichments to access derived contexts — see e.g. https://discourse.snowplow.io/t/custom-javascript-enrichment-uses-cookie-info/5579

stanch:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -662,9 +662,12 @@ object EnrichmentManager {
...
   ): EStateT[F, Unit] =
     EStateT.fromEither {
-      case (event, _) =>
+      case (event, derivedContexts) =>
         javascriptScript match {
-          case Some(jse) => jse.process(event).leftMap(NonEmptyList.one(_))
+          case Some(jse) =>
+            if (derivedContexts.nonEmpty)
+              event.derived_contexts = ME.formatDerivedContexts(derivedContexts)
+            jse.process(event).leftMap(NonEmptyList.one(_))
```
stanch, line 670: Note: that still does not enable `event.setDerived_contexts()`, however I think we can make a reasonable argument that it’s not necessary.

benjben (commented):
LGTM!

It'd be nice to add a unit test where a derived context is used in JS enrichment.

benjben (approved):
:+1:

PR #787, opened 2023-05-02
Deprecate Stream Enrich and enrich-rabbitmq

spenes:


modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/enrichments/registry/pii/PiiPseudonymizerEnrichmentSpec.scala
```diff
@@ -199,7 +202,7 @@ class PiiPseudonymizerEnrichmentSpec extends Specification with ValidatedMatcher
       "jsonschema",
       SchemaVer.Full(2, 0, 0)
     )
-    IpLookupsEnrichment.parse(js, schemaKey, true).toOption.get.enrichment[Id](BlockerF.noop)
+    IpLookupsEnrichment.parse(js, schemaKey, true).toOption.get.enrichment[IO](BlockerF.noop).unsafeRunSync()
```
benjben, line 205:
```suggestion
    IpLookupsEnrichment.parse(js, schemaKey, true).toOption.get.enrichment[IO](BlockerF.noop)
```

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/utils/Clock.scala
```diff
@@ -13,16 +13,8 @@
...
+import cats.effect.{Clock => CEClock, IO}
 
 object Clock {
-  implicit val idClock: CEClock[Id] = new CEClock[Id] {
-    final def realTime(unit: TimeUnit): Id[Long] =
-      unit.convert(System.currentTimeMillis(), TimeUnit.MILLISECONDS)
-    final def monotonic(unit: TimeUnit): Id[Long] =
-      unit.convert(System.nanoTime(), TimeUnit.NANOSECONDS)
-  }
+  implicit val ioClock: CEClock[IO] = CEClock.create[IO]
```
benjben, line 19: Can probably be removed

benjben (changes requested):
Great work @spenes !

I wonder if we should merge `common/` and `common-fs2`. On one hand it's probably good to keep `common/` as is with the just the core logic of the adapters/enrichments. But on the other hand it sounds weird to have two modules with some common logic. Given that Scala Common Enrich is published as a library to Maven in case someone wants to use it in a custom app, I think that we should leave it as is.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/SpecHelpers.scala
```diff
@@ -63,11 +72,16 @@ object SpecHelpers {
   }"""
 
   /** Builds an Iglu client from the above Iglu configuration. */
-  val client: IgluCirceClient[Id] = IgluCirceClient
-    .parseDefault[Id](igluConfig)
+  val client: IgluCirceClient[IO] = IgluCirceClient
+    .parseDefault[IO](igluConfig)
     .value
+    .unsafeRunSync()
```
benjben, line 78: I don't like that but changing this would require *a lot* of refactoring in the tests. Instead I would prefer us to make it possible to create an `IgluCirceClient` manually without having to parse a config file, so that we don't get a `IO[ ]`.

spenes (commented):
Looks good to me! However, Github doesn't allow me to approve the PR since I am the one opened the PR. But, from my point of view, it is good to go.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
```diff
@@ -91,20 +90,26 @@ class AdapterRegistry(
...
+      case Some(adapter) =>
+        adapter.toRawEvents(payload, client)
       case None =>
-        val f = FailureDetails.AdapterFailure.InputData(
-          "vendor/version",
-          Some(s"${payload.api.vendor}/${payload.api.version}"),
-          "vendor/version combination is not supported"
-        )
-        Monad[F].pure(f.invalidNel[NonEmptyList[RawEvent]])
+        remoteAdapters.get((payload.api.vendor, payload.api.version)) match {
```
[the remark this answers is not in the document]
benjben: `adapters.get` is used above, and if there is no result in the standard adapters, we then check in the remote ones

PR #790, opened 2023-05-03
Draft currency conversion configuration

istreeter:

### DIFF
--- a/config/enrichments/currency_conversion_config.json
+++ b/config/enrichments/currency_conversion_config.json
@@ -1,6 +1,5 @@
 {
-	"schema": "iglu:com.snowplowanalytics.snowplow/currency_conversion_config/jsonschema/1-0-0",
-
+	"schema": "iglu:com.snowplowanalytics.snowplow/currency_conversion_config/jsonschema/2-0-0",
 	"data": {
 
 		"enabled": false,
@@ -11,6 +10,50 @@
 			"apiKey": "{{KEY}}",
 			"baseCurrency": "USD",
 			"rateAt": "EOD_PRIOR"
+		},
+		"mappings": [
+		{
+			"input": "iglu:ecommerce.tracking/product_entity/jsonschema/1-0-0",
+			"output": "iglu:ecommerce.tracking/converted_product_entity/jsonschema/1-0-0",
+			"fields": [
+			{
+				"inputField": "price",
+				"currency": "currency",
+				"outputField": "price"
+			},
+			{
+				"inputField": "list_price",
+				"currency": "currency",
+				"outputField": "list_price"
+			}
+			]
+		},
+		{
+			"input": "iglu:ecommerce.tracking/transaction_entity/jsonschema/1-0-0",
+			"output": "iglu:ecommerce.tracking/converted_transaction_entity/jsonschema/1-0-0",
+			"fields": [
+			{
+				"inputField": "revenue",
+				"currency": "currency",
+				"outputField": "revenue"
+			},
+			{
+				"inputField": "tax",
+				"currency": "currency",
+				"outputField": "tax"
+			},
+			{
+				"inputField": "shipping",
+				"currency": "currency",
+				"outputField": "shipping"
+			},
+			{
+				"inputField": "discount_amount",
+				"currency": "currency",
+				"outputField": "discount_amount"
+			}
+			]
 		}
+		]
 	}
 }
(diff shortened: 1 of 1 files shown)


config/enrichments/currency_conversion_config.json
```diff
@@ -11,6 +10,50 @@
 			"apiKey": "{{KEY}}",
 			"baseCurrency": "USD",
 			"rateAt": "EOD_PRIOR"
+		},
+		"mappings": [
+		{
+			"input": "iglu:ecommerce.tracking/product_entity/jsonschema/1-0-0",
```
istreeter, line 16: Do we need a field here to say whether to look in the contexts vs in the unstruct_event?

miike: Yes - I think that'd be a good call.

config/enrichments/currency_conversion_config.json
```diff
@@ -11,6 +10,50 @@
...
+			"output": "iglu:ecommerce.tracking/converted_product_entity/jsonschema/1-0-0",
+			"fields": [
+			{
+				"inputField": "price",
+				"currency": "currency",
+				"outputField": "price"
+			},
+			{
+				"inputField": "list_price",
+				"currency": "currency",
```
miike, line 26: Silly question - is this the target currency that we are converting to or something else?

config/enrichments/currency_conversion_config.json
```diff
@@ -11,6 +10,50 @@
 			"apiKey": "{{KEY}}",
 			"baseCurrency": "USD",
 			"rateAt": "EOD_PRIOR"
```
miike, line 12: @istreeter What's your opinion on exposing the caching settings in parameters here similar to what we do for the API enrichment?

config/enrichments/currency_conversion_config.json
istreeter: It says which field from the `product_entity` schema contains the relevant currency.  Imagine the incoming context looks like this:

```
{
  "price": 42.00,
  "list_price": 42.00,
  "currency": "USD"
}
```

config/enrichments/currency_conversion_config.json
istreeter: Something like `"entity": "event"` vs `"entity": "context"`?  I hate naming things!

config/enrichments/currency_conversion_config.json
istreeter: Is there ever a chance that the incoming entity might contain two different currency fields?  If no, then I could move this currency setting up a level in the config.  So there is just a single currency per mapping.

config/enrichments/currency_conversion_config.json
benjben: [This](https://github.com/snowplow/enrich/blob/master/config/enrichments/sql_query_enrichment_config.json#L20) is how it's done in other enrichments

istreeter: Oh, I should have checked what we do in other places.  I'll update this to be consistent with the other configs.  Thanks!

config/enrichments/currency_conversion_config.json
miike: Within the same entity as part of the same event? Probably pretty unlikely as most ecomm won't support multi-currency checkouts.

config/enrichments/currency_conversion_config.json
istreeter: I paused for ages on this question!  It makes me nervous to expose such a powerful setting in such a user-facing piece of config.  Cache size is a balance between performance vs available memory, and the person configuring this enrichment might not know all of the context for tuning that.

On balance, though, it's better to make it configurable because we never know when we'll need to tune things at run time.  So yeah we should put it in this config.

@miike what's your opinion on the `ignoreOnError` setting that [is newly added to the sql enrichment config](https://github.com/snowplow/iglu-central/blob/master/schemas/com.snowplowanalytics.snowplow.enrichments/sql_query_enrichment_config/jsonschema/1-0-1#L204-L206)?  If we added it here, it would basically mean.... if we fail to fetch a rate because of a network error, then it's OK to emit the event without the derived context.

miike: I like the idea of `ignoreOnError` as there are going to be circumstances where a failed enrichment event (e.g., on checkout) is going to be annoying to not have because a currency conversion couldn't be performed.

I think this does raise the question (for both enrichments) about how you "recover" data that hasn't generated a failed event but might be NULL in the data warehouse though. I don't know that we can (or want to?) change our philosophy around the immutability of these events.

PR #792, opened 2023-05-18
Make schemas configurable in adapters (close #791)

### LINKED ISSUE #791: SendGrid webhook: make schema configurable
## Context

`network/json-schema-validator` library [uses a manually-maintained list of TLDs to validate the emails](https://github.com/networknt/json-schema-validator/blob/5a67ecc1171ed59ac6febdff247003b5ae37ea75/src/main/java/com/networknt/schema/format/DomainValidator.java#L290), which can cause the emails sent via SendGrid webhook to fail validation because their domain is not known.

A possible solution is to make the schema [hard coded in Sendgrid adapter](https://github.com/snowplow/enrich/blob/3.8.0/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/SendgridAdapter.scala#L43-L45) configurable.

## Guide

Enrich has kind of 3 steps:

- [loaders](https://github.com/snowplow/enrich/blob/3.8.0/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/loaders): deserialize the Thrift payload
- [adapters](https://github.com/snowplow/enrich/blob/3.8.0/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters): take the info from the HTTP request (~ collector payload) to create a unified [CollectorPayload](https://github.com/snowplow/enrich/blob/3.8.0r/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/loaders/CollectorPayload.scala#L52-L59), that has the same format wherever the source (webhook, Snowplow event..)
- [enrichments](https://github.com/snowplow/enrich/blob/3.8.0/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments): this is where the validation of the input event happens, as well as the actual enrichments

The configuration format (for example for [Kinesis](https://github.com/snowplow/enrich/blob/3.8.0/config/config.kinesis.extended.hocon)) is defined in [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala#L50). We probably want to add a new section that could look like this :

```
"adapters": {
  "sendgrid": {
    "schema": "iglu://..."
  }
}
```

There are [several sbt modules](https://github.com/snowplow/enrich/blob/3.8.0/modules):

- `common` contains the adapters and enrichments logic (we'll need to update it so that the adapters can become configurable)
- `common-fs2` contains the logic related to configuring the app, reading events, writing them (we'll need to update it so that the config is propagated to the adapters)
- `kinesis`: logic related to reading from/writing to Kinesis (we won't need to touch it)
- ...

When running `enrich-kinesis` for instance, the starting point is [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/Main.scala#L40).
It then goes [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/kinesis/src/main/scala/com/snowplowanalytics/snowplow/enrich/kinesis/KinesisRun.scala#L43) and then goes to `common-fs2`'s `Run` [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala#L42).
It creates an `Environment` [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Run.scala#L87).
In our case we'll need to take the schema defined in the config and pass it to the `AdapterRegistry` [here](https://github.com/snowplow/enrich/blob/3.8.0/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala#L201) when creating the `Environment`

matus-tomlein:
Issue #791 

This PR adds configuration for adapters to change the schema URIs used for tracking events and entities. The goal is to make it possible for the user to change the schema without having to redeploy the adapter, which would for example enable them to change validation rules of properties in the schemas.

The suggested approach is quite general and enables changing the schemas in any adapter (except the `CloudfrontAccessLogAdapter` which tracks multiple versions of the same event schema). One configures the schemas by adding a dictionary under `adapters.schemas` in the config file. This dictionary first maps schema vendors to sub-dictionaries with a mapping of schema names to new schema URIs. For example:

```hocon
{
  ...

  "adapters": {
    # This would make any adapter tracking an event using a schema such as `iglu:com.acme/event/1-0-0`
    # to use `iglu:com.custom/other_event/jsonschema/1-2-3` instead.
    "schemas": {
      "com.acme": {
        "event": "iglu:com.custom/other_event/jsonschema/1-2-3"
      }
    }
  }
}
```

To implement this, I changed the adapters to be `class` instead of `object` and accept a `config: AdapterConfig` that contains the `schemas` mapping. The adapters instead of calling `SchemaKey(vendor, name, format, version)` to create a schema URI, now call `config.schemaKey(vendor, name, format, version)`.

### DIFF
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
@@ -198,7 +198,7 @@ object Environment {
       metadata <- Resource.eval(metadataReporter[F](file, processor.artifact, http))
       assets = parsedConfigs.enrichmentConfigs.flatMap(_.filesToCache)
       (remoteAdaptersHttpClient, remoteAdapters) <- prepareRemoteAdapters[F](file.remoteAdapters, ec, metrics)
-      adapterRegistry = new AdapterRegistry(remoteAdapters)
+      adapterRegistry = new AdapterRegistry(remoteAdapters, file.adaptersSchemas)
       sem <- Resource.eval(Semaphore(1L))
       assetsState <- Resource.eval(Assets.State.make[F](blocker, sem, clts, assets))
       shifter <- ShiftExecution.ofSingleThread[F]
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala
@@ -22,19 +22,10 @@ import _root_.io.circe.{Decoder, Encoder}
 import _root_.io.circe.config.syntax._
 import _root_.io.circe.generic.extras.semiauto.{deriveConfiguredDecoder, deriveConfiguredEncoder}
 
-import com.typesafe.config.{Config => TSConfig, ConfigFactory}
+import com.typesafe.config.{ConfigFactory, Config => TSConfig}
 
-import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.{
-  Concurrency,
-  Experimental,
-  FeatureFlags,
-  Input,
-  Monitoring,
-  Output,
-  Outputs,
-  RemoteAdapterConfigs,
-  Telemetry
-}
+import com.snowplowanalytics.snowplow.enrich.common.adapters.AdaptersSchemas
+import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io._
 
 /**
  * Parsed HOCON configuration file
@@ -46,6 +37,7 @@ import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.{
  * @param telemetry configuration for telemetry
  * @param featureFlags to activate/deactivate enrich features
  * @param experimental configuration for experimental features
+ * @param adaptersSchemas configuration for adapters
  */
 final case class ConfigFile(
   input: Input,
@@ -56,10 +48,12 @@ final case class ConfigFile(
   monitoring: Monitoring,
   telemetry: Telemetry,
   featureFlags: FeatureFlags,
-  experimental: Option[Experimental]
+  experimental: Option[Experimental],
+  adaptersSchemas: AdaptersSchemas
 )
 
 object ConfigFile {
+  import AdaptersSchemasEncoderDecoders._
 
   // Missing in circe-config
   implicit val finiteDurationEncoder: Encoder[FiniteDuration] =
@@ -67,16 +61,16 @@ object ConfigFile {
 
   implicit val configFileDecoder: Decoder[ConfigFile] =
     deriveConfiguredDecoder[ConfigFile].emap {
-      case ConfigFile(_, _, _, Some(aup), _, _, _, _, _) if aup._1 <= 0L =>
+      case ConfigFile(_, _, _, Some(aup), _, _, _, _, _, _) if aup._1 <= 0L =>
         "assetsUpdatePeriod in config file cannot be less than 0".asLeft // TODO: use newtype
       // Remove pii output if streamName and region empty
-      case c @ ConfigFile(_, Outputs(good, Some(output: Output.Kinesis), bad), _, _, _, _, _, _, _) if output.streamName.isEmpty =>
+      case c @ ConfigFile(_, Outputs(good, Some(output: Output.Kinesis), bad), _, _, _, _, _, _, _, _) if output.streamName.isEmpty =>
         c.copy(output = Outputs(good, None, bad)).asRight
       // Remove pii output if topic empty
-      case c @ ConfigFile(_, Outputs(good, Some(Output.PubSub(t, _, _, _, _)), bad), _, _, _, _, _, _, _) if t.isEmpty =>
+      case c @ ConfigFile(_, Outputs(good, Some(Output.PubSub(t, _, _, _, _)), bad), _, _, _, _, _, _, _, _) if t.isEmpty =>
         c.copy(output = Outputs(good, None, bad)).asRight
       // Remove pii output if topic empty
-      case c @ ConfigFile(_, Outputs(good, Some(Output.Kafka(topicName, _, _, _, _)), bad), _, _, _, _, _, _, _) if topicName.isEmpty =>
+      case c @ ConfigFile(_, Outputs(good, Some(Output.Kafka(topicName, _, _, _, _)), bad), _, _, _, _, _, _, _, _) if topicName.isEmpty =>
         c.copy(output = Outputs(good, None, bad)).asRight
       case other => other.asRight
     }
@@ -88,9 +82,12 @@ object ConfigFile {
    * Command line parameters have the highest priority.
    * Then the provided file.
    * Then the application.conf file.
+   * Then the reference.conf file.
    */
-  private def configFileFallbacks(in: TSConfig): TSConfig =
-    ConfigFactory.load(in.withFallback(ConfigFactory.load()))
+  private def configFileFallbacks(in: TSConfig): TSConfig = {
+    val defaultConf = ConfigFactory.load("reference.conf")
+    ConfigFactory.load(in.withFallback(ConfigFactory.load().withFallback(defaultConf)))
+  }
 
   def parse[F[_]: Sync](raw: EncodedHoconOrPath): EitherT[F, String, ConfigFile] =
     ParsedConfigs.parseEncodedOrPath(raw, configFileFallbacks)
--- a/modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
+++ b/modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
@@ -31,6 +31,7 @@ import org.http4s.Uri
 import org.specs2.mutable.Specification
 
 import com.snowplowanalytics.snowplow.enrich.common.fs2.config.io.BackoffPolicy
+import com.snowplowanalytics.snowplow.enrich.common.SpecHelpers.adaptersSchemas
 
 class ConfigFileSpec extends Specification with CatsIO {
   "parse" should {
@@ -90,7 +91,8 @@ class ConfigFileSpec extends Specification with CatsIO {
               )
             )
           )
-        )
+        ),
+        adaptersSchemas
       )
       ConfigFile.parse[IO](configPath.asRight).value.map(result => result must beRight(expected))
     }
@@ -191,7 +193,8 @@ class ConfigFileSpec extends Specification with CatsIO {
               )
             )
           )
-        )
+        ),
+        adaptersSchemas
       )
       ConfigFile.parse[IO](configPath.asRight).value.map(result => result must beRight(expected))
     }
@@ -302,7 +305,8 @@ class ConfigFileSpec extends Specification with CatsIO {
               )
             )
           )
-        )
+        ),
+        adaptersSchemas
       )
       ConfigFile.parse[IO](configPath.asRight).value.map(result => result must beRight(expected))
     }
@@ -403,7 +407,8 @@ class ConfigFileSpec extends Specification with CatsIO {
               )
             )
           )
-        )
+        ),
+        adaptersSchemas
       )
       ConfigFile.parse[IO](configPath.asRight).value.map(result => result must beRight(expected))
     }
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
@@ -34,29 +34,31 @@ import com.snowplowanalytics.snowplow.enrich.common.utils.HttpClient
  * The AdapterRegistry lets us convert a CollectorPayload into one or more RawEvents, using a given
  * adapter.
  */
-class AdapterRegistry(remoteAdapters: Map[(String, String), RemoteAdapter] = Map.empty) {
-
+class AdapterRegistry(
+  remoteAdapters: Map[(String, String), RemoteAdapter] = Map.empty,
+  adaptersSchemas: AdaptersSchemas
+) {
   val adapters: Map[(String, String), Adapter] = Map(
     (Vendor.Snowplow, "tp1") -> Tp1Adapter,
     (Vendor.Snowplow, "tp2") -> Tp2Adapter,
     (Vendor.Redirect, "tp2") -> RedirectAdapter,
     (Vendor.Iglu, "v1") -> IgluAdapter,
-    (Vendor.Callrail, "v1") -> CallrailAdapter,
-    (Vendor.Cloudfront, "wd_access_log") -> CloudfrontAccessLogAdapter,
-    (Vendor.Mailchimp, "v1") -> MailchimpAdapter,
-    (Vendor.Mailgun, "v1") -> MailgunAdapter,
-    (Vendor.GoogleAnalytics, "v1") -> GoogleAnalyticsAdapter,
-    (Vendor.Mandrill, "v1") -> MandrillAdapter,
-    (Vendor.Olark, "v1") -> OlarkAdapter,
-    (Vendor.Pagerduty, "v1") -> PagerdutyAdapter,
-    (Vendor.Pingdom, "v1") -> PingdomAdapter,
-    (Vendor.Sendgrid, "v3") -> SendgridAdapter,
-    (Vendor.StatusGator, "v1") -> StatusGatorAdapter,
-    (Vendor.Unbounce, "v1") -> UnbounceAdapter,
-    (Vendor.UrbanAirship, "v1") -> UrbanAirshipAdapter,
-    (Vendor.Marketo, "v1") -> MarketoAdapter,
-    (Vendor.Vero, "v1") -> VeroAdapter,
-    (Vendor.HubSpot, "v1") -> HubSpotAdapter
+    (Vendor.Callrail, "v1") -> CallrailAdapter(adaptersSchemas.callrail),
+    (Vendor.Cloudfront, "wd_access_log") -> CloudfrontAccessLogAdapter(adaptersSchemas.cloudfrontAccessLog),
+    (Vendor.Mailchimp, "v1") -> MailchimpAdapter(adaptersSchemas.mailchimp),
+    (Vendor.Mailgun, "v1") -> MailgunAdapter(adaptersSchemas.mailgun),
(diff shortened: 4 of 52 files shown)


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package com.snowplowanalytics.snowplow.enrich.common.adapters
+
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer}
+
+/**
+ * Configuration for an adapter.
+ * @param schemas optional map of vendor name to event name to schema URI
+ */
+case class AdapterConfig(schemas: Option[Map[String, Map[String, String]]] = None) {
```
benjben, line 22: Before your changes, we were always using hard coded schemas in the adapters. Now we make it possible to override them. The way we usually do this is to remove entirely the hard coded values from the code, and put all the defaults in `src/main/resources/application.conf`.

This could look something like
```
"adapters": {
  "callrail": {
    "call_complete": "iglu:com.callrail/call_complete/jsonschema/1-0-2"
  },
  "googleAnalytics": {
    ...,
    "screenview": "iglu:com.google.analytics/screenview/jsonschema/1-0-0",
    "transaction": "iglu:com.google.analytics/transaction/jsonschema/1-0-0"
  },
  ...
}
````
The main difficulty with this approach is that at the moment the defaults are only per app (e.g. for [enrich-kinesis](https://github.com/snowplow/enrich/blob/master/modules/kinesis/src/main/resources/application.conf)), there is no default in [common/src/resources/](https://github.com/snowplow/enrich/tree/master/modules/common/src), so we'd need to add one there and make sure that it is resolved by the apps when loading the configuration.

If this is too complicated, I think that we should at least organize the schemas by adapter in the config, e.g.
```
"adapters": {
  "callrail": {
    "schemas": {
      { "com.callrail" : { "call_complete": "iglu:com.callrail/call_complete/jsonschema/1-0-2" }}
    }
  },
  "googleAnalytics" : {
    "schemas": {
      { "com.google.analytics": {
        "event": "iglu:com.google.analytics/event/jsonschema/1-0-0" },
        "screen_view": "iglu:com.google.analytics/screen_view/jsonschema/1-0-0"}
      }
    }
  }
}
```
and then pass the appropriate schemas when instantating an adapter, e.g.
```
new CallrailAdapter(adapterConfig.callrail)
```

**Review by benjben — CHANGES_REQUESTED** [review]
Hi @matus-tomlein ,

The changes look great and would work !

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala (hunk 1 above)
Comment (matus-tomlein) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala]: Hi @benjben, thanks for the feedback!

I have changed it to only store the defaults in the configuration and added a `application-default.conf` file to `common/src/resources` with the default values.

The problem is that this resulted in quite a bit of duplication of the configuration:

1. the values are now present in all the `extended` config files.
2. I had to make them also available in the tests, which I did by adding them to the `SpecHelpers`. However since there is a different `SpecHelpers` for each test module, I had to duplicate this too.

If you have any suggestions or ideas how to prevent the duplication, I would be happy to change it. I especially don't like the duplication in the `SpecHelpers` but couldn't find a different way to do it.

Comment (matus-tomlein) [reply modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala]: I have made the changes as we discussed (remove the config from extended conf files and share the same `SpecHelpers` defaults) and it helped a lot with the config duplication!

File: modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala [hunk 2]
```diff
@@ -0,0 +1,34 @@
...
+  def schemaKey(
+    vendor: String,
+    name: String,
+    format: String,
+    version: SchemaVer.Full
+  ): SchemaKey =
+    schemas
+      .flatMap(_.get(vendor))
+      .flatMap(_.get(name))
+      .flatMap(uri => SchemaKey.fromUri(uri).toOption)
```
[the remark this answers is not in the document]
matus-tomlein: Have changed the way this works, but now it throws errors in case the config is not present.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterConfig.scala
```diff
@@ -0,0 +1,34 @@
...
+package com.snowplowanalytics.snowplow.enrich.common.adapters
+
+import com.snowplowanalytics.iglu.core.{SchemaKey, SchemaVer}
+
+/**
+ * Configuration for an adapter.
+ * @param schemas optional map of vendor name to event name to schema URI
+ */
+case class AdapterConfig(schemas: Option[Map[String, Map[String, String]]] = None) {
+  def schemaKey(
```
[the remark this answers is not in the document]
matus-tomlein: That sounds better, but I have removed the function in the last changes.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/adapters/registry/GoogleAnalyticsAdapterSpec.scala
```diff
@@ -52,6 +51,7 @@ class GoogleAnalyticsAdapterSpec extends Specification with DataTables with Vali
   breakDownCompositeField should work properly                             $e20
   """
 
+  val AdapterWithoutConfig = new GoogleAnalyticsAdapter()
```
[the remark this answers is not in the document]
matus-tomlein: Have tried to make this consistent!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -13,34 +13,26 @@
...
-
 import com.snowplowanalytics.snowplow.badrows.Processor
 import com.snowplowanalytics.snowplow.enrich.common.adapters.AdapterRegistry
 import com.snowplowanalytics.snowplow.enrich.common.adapters.registry.RemoteAdapter
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.EnrichmentRegistry
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.registry.EnrichmentConf
 import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent
 import com.snowplowanalytics.snowplow.enrich.common.utils.{BlockerF, HttpClient, ShiftExecution}
-
 import com.snowplowanalytics.snowplow.enrich.common.fs2.config.{ConfigFile, ParsedConfigs}
```
benjben, line 36: Do you have a formatter set up ? We ususally like it to separate the packages by lib with an empty line in-between

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
```diff
@@ -198,7 +190,25 @@ object Environment {
       metadata <- Resource.eval(metadataReporter[F](file, processor.artifact, http))
       assets = parsedConfigs.enrichmentConfigs.flatMap(_.filesToCache)
       (remoteAdaptersHttpClient, remoteAdapters) <- prepareRemoteAdapters[F](file.remoteAdapters, ec, metrics)
-      adapterRegistry = new AdapterRegistry(remoteAdapters)
+      adapterRegistry = new AdapterRegistry(
+                          remoteAdapters,
+                          callrailConfig = file.adapters.callrail,
```
benjben, line 195:
```suggestion
      adapterRegistry = new AdapterRegistry(remoteAdapters, file.adapters)
```
It'd be better to let `AdapterRegistry` pick the configs

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFile.scala
```diff
@@ -88,9 +87,12 @@ object ConfigFile {
    * Command line parameters have the highest priority.
    * Then the provided file.
    * Then the application.conf file.
+   * Then the application-default.conf file.
    */
-  private def configFileFallbacks(in: TSConfig): TSConfig =
-    ConfigFactory.load(in.withFallback(ConfigFactory.load()))
+  private def configFileFallbacks(in: TSConfig): TSConfig = {
+    val defaultConf = ConfigFactory.load("application-default.conf")
```
benjben, line 93:
```suggestion
    val defaultConf = ConfigFactory.load("application.conf")
```
To be consistent with the rest of the codebase (although it should better be names `reference.conf`)

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
```diff
@@ -15,21 +15,15 @@ package com.snowplowanalytics.snowplow.enrich.common.fs2.config
...
-
 import scala.concurrent.duration._
-
 import cats.syntax.either._
 import cats.data.NonEmptyList
 import cats.effect.IO
-
 import cats.effect.testing.specs2.CatsIO
-
+import com.snowplowanalytics.snowplow.enrich.common.SpecHelpers._
```
benjben, line 23: We usually put Snowplow packages at the bottom of the import

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
```diff
@@ -90,6 +84,24 @@ class ConfigFileSpec extends Specification with CatsIO {
...
+          mandrill = mandrillConfig,
+          marketo = marketoConfig,
+          olark = olarkConfig,
+          pagerduty = pagerdutyConfig,
+          pingdom = pingdomConfig,
+          sendgrid = sendgridConfig,
+          statusgator = statusGatorConfig,
+          unbounce = unbounceConfig,
+          urbanAirship = urbanAirshipConfig,
+          vero = veroConfig
```
benjben, line 104: You can define this once somewhere and reuse it everywhere

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
```diff
@@ -34,29 +34,46 @@ import com.snowplowanalytics.snowplow.enrich.common.utils.HttpClient
...
+  marketoConfig: MarketoConfig,
+  olarkConfig: OlarkConfig,
+  pagerdutyConfig: PagerdutyConfig,
+  pingdomConfig: PingdomConfig,
+  sendgridConfig: SendgridConfig,
+  statusGatorConfig: StatusGatorConfig,
+  unbounceConfig: UnbounceConfig,
+  urbanAirshipConfig: UrbanAirshipConfig,
+  veroConfig: VeroConfig
+) {
```
benjben, line 55:
```suggestion
class AdapterRegistry(
  remoteAdapters: Map[(String, String), RemoteAdapter] = Map.empty,
  schemas: AdaptersSchemas
) {
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/CallrailAdapter.scala
```diff
@@ -37,19 +32,14 @@ import Adapter.Adapted
  * a known version of the AD-X Tracking webhook
  * into raw events.
  */
-object CallrailAdapter extends Adapter {
+class CallrailAdapter(config: CallrailConfig) extends Adapter {
```
benjben, line 35:
```suggestion
case class CallrailAdapter(config: CallrailConfig) extends Adapter {
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
```diff
@@ -34,29 +34,46 @@ import com.snowplowanalytics.snowplow.enrich.common.utils.HttpClient
...
-    (Vendor.Pagerduty, "v1") -> PagerdutyAdapter,
-    (Vendor.Pingdom, "v1") -> PingdomAdapter,
-    (Vendor.Sendgrid, "v3") -> SendgridAdapter,
-    (Vendor.StatusGator, "v1") -> StatusGatorAdapter,
-    (Vendor.Unbounce, "v1") -> UnbounceAdapter,
-    (Vendor.UrbanAirship, "v1") -> UrbanAirshipAdapter,
-    (Vendor.Marketo, "v1") -> MarketoAdapter,
-    (Vendor.Vero, "v1") -> VeroAdapter,
-    (Vendor.HubSpot, "v1") -> HubSpotAdapter
+    (Vendor.Callrail, "v1") -> new CallrailAdapter(callrailConfig),
```
benjben, line 61: Using `case class` instead of `class` will remove the need to use `new` and will be closer to the `object` that was there before

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/adapters/registry/CallrailAdapterSpec.scala
```diff
@@ -34,6 +34,8 @@ class CallrailAdapterSpec extends Specification with DataTables with ValidatedMa
   toRawEvents should return a Validation Failure if there are no parameters on the querystring      $e2
   """
 
+  val AdapterWithDefaultConfig = new CallrailAdapter(config = callrailConfig)
```
benjben, line 37:
```suggestion
  val adapterWithDefaultConfig = new CallrailAdapter(config = callrailConfig)
```
We generally use lowerCamelCase in the codebase

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/adapters/registry/GoogleAnalyticsAdapterSpec.scala
```diff
@@ -481,23 +481,24 @@ class GoogleAnalyticsAdapterSpec extends Specification with DataTables with Vali
         "composite field name has to conform to regex " +
           """(pr|promo|il|cd|cm|cg)(\d+)([a-zA-Z]*)(\d*)([a-zA-Z]*)(\d*)$"""
       )
+    val adapter = AdapterWithDefaultConfig
```
benjben, line 484: Useless

benjben (changes requested):

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
matus-tomlein: Sorry I didn't realize there was a convention for this, my IDE had a setting to reorder the imports alphabeticallly. Have reverted it now!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Environment.scala
matus-tomlein: Have changed this to accept a `AdaptersSchemas` instance with all the configs encapsulated.

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
matus-tomlein: Have tried to follow this convention too.

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/ConfigFileSpec.scala
matus-tomlein: Have defined `adaptersSchemas` in `SpecHelpers` to be reused here.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
matus-tomlein: Done! Also good idea on renaming `AdaptersConfig` to `AdaptersSchemas`. Have also changed this in the configuration files to use the key `adaptersSchemas` instead of `adapters` since that only configures the schemas for adapters.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/registry/CallrailAdapter.scala
matus-tomlein: Have changed all adapters to `case class`

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/adapters/AdapterRegistry.scala
matus-tomlein: Done!

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/adapters/registry/CallrailAdapterSpec.scala
matus-tomlein: Changed.

modules/common/src/test/scala/com.snowplowanalytics.snowplow.enrich.common/adapters/registry/GoogleAnalyticsAdapterSpec.scala
matus-tomlein: Removed.

benjben (approved):
It looks great @matus-tomlein thank you!

Have you tested the feature by running the Docker image and using a config file?

matus-tomlein:
Thanks @benjben! I was testing it by building a jar file (`sbt 'project kinesis' assembly`) and then running the jar with custom configuration where I had some modified schemas under `adapterSchemas`. Is there a different way how I should test it?

Also what is the process for merging this PR? Should I merge it to a release branch?

Issue #761, opened 2023-03-31
Security vulnerability on snowplow-stream-enrich-kinesis-3.7.0.jar


VISHAL-PERIYASAMY:
**Jar**:  snowplow-stream-enrich-kinesis

**Version**: 3.7.0

**Security Issue CVE**: 
1. CVE-2022-41854
2. CVE-2022-38752
3. CVE-2022-1471
4. CVE-2023-1370
5. CVE-2022-31159
6. CVE-2021-29425

**Security Issue Package Fix**:
1. snowplow-stream-enrich-kinesis-3.7.0.jar (3.7.0): org.yaml:snakeyaml @ 2.0
2. snowplow-stream-enrich-kinesis-3.7.0.jar (3.7.0): net.minidev:json-smart @ 2.4.9
3. snowplow-stream-enrich-kinesis-3.7.0.jar (3.7.0): com.amazonaws:aws-java-sdk-s3 @ 1.12.261
4. snowplow-stream-enrich-kinesis-3.7.0.jar (3.7.0): commons-io:commons-io @ 2.7

VISHAL-PERIYASAMY:
Hi team, any update on this particular issue?
Even on the 3.7.1 version also the security issues are not fixed.

benjben:
Hi @VISHAL-PERIYASAMY ,
`snowplow-stream-enrich-kinesis` will get deprecated in the next weeks.
It's been replaced by `enrich-kinesis`, which we actively support and has no critical and high vulnerabilities.
You can find the info to set it up on our [docs website](https://docs.snowplow.io/docs/pipeline-components-and-applications/enrichment-components/enrich-kinesis/).

VISHAL-PERIYASAMY:
Hi @benjben 

The Security issues are not resolved on the snowplow-enrich-kinesis-3.7.2.jar

**Critical issue:**
- CVE-2022-1471

**Medium Issue:**
- CVE-2022-41854
- CVE-2022-38752
- CVE-2022-31159
- CVE-2021-29425

benjben:
Hi @VISHAL-PERIYASAMY ,

We're committed to remove all critical and high vulnerabilities in our apps.

The CVEs that you're listing appear as medium or low in [Snyk](https://snyk.io/).

Feel free to open a PR with these libs bumped if it's important for you to have them removed.

VISHAL-PERIYASAMY:
Sure @benjben  looks like I am not able to create a branch/ push my changes.
Can you guide me on it?

[...]

PR #800, opened 2023-07-17
Add info to observed_event

benjben:


modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/GcsClient.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import blobstore.gcs.GcsStore
+import blobstore.url.Url
+
+import com.google.cloud.storage.StorageOptions
+import com.google.cloud.BaseServiceException
+
+import com.snowplowanalytics.snowplow.enrich.common.fs2.io.Clients.{Client, RetryableFailure}
+
+object GcsClient {
```
istreeter, line 30: Is there an argument this should live in the common-fs2 module instead of repeating it in the stream-specific modules?  common-fs2 already [has dependencies on the s3 and gcs clients](https://github.com/snowplow/enrich/blob/3.8.0/project/Dependencies.scala#L280-L281)

modules/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.kafka/Main.scala
```diff
@@ -60,7 +61,7 @@ object Main extends IOApp.WithContext {
       List.empty,
       _.record.value,
       MaxRecordSize,
-      None,
+      Some(Cloud.Azure),
```
stanch, line 64: This could be a bit misleading... since someone can run Enrich Kafka outside of Azure

benjben: Yeah I thought about that. But if we don't do that we need to make the cloud configurable. Given that this is only for telemetry for now, and that we are not even sure to keep it (unless I missed something), I assumed that this was the best call. Do you think that I should change that?

stanch: Is there anything in the Kafka config for Azure that we could use to identify that we are running there? I know we need to set these properties, but they seem a bit generic...

```
// Event Hubs specific settings using a connection string
props.setProperty(“security.protocol”, “SASL_SSL”)
props.setProperty(“sasl.mechanism”, “PLAIN”)
props.setProperty(“sasl.jaas.config”, “”“org.apache.kafka.common.security.plain.PlainLoginModule required username=”$ConnectionString" password=“<your_connection_string>”;“”")

// Event Hubs specific settings for OAuth
props.setProperty(“security.protocol”, “SASL_SSL”)
props.setProperty(“sasl.mechanism”, “OAUTHBEARER”)
props.setProperty(“sasl.jaas.config”, “org.apache.kafka.common.security.oauthbearer.OAuthBearerLoginModule required;”)
props.setProperty(“sasl.login.callback.handler.class”, “CustomAuthenticateCallbackHandler”)
```

benjben: I'm afraid that there isn't anything we can use, as these security parameters are generic and can be used outside of Azure

project/Dependencies.scala
```diff
@@ -34,8 +34,8 @@ object Dependencies {
     val commonsText      = "1.10.0"
     val jodaTime         = "2.10.1"
     val useragent        = "1.21"
-    val uaParser         = "1.4.3"
-    val snakeYaml        = "1.31" // override transitive dependency to mitigate security vulnerabilities CVE-2022-25857
+    val uaParser         = "1.5.4"
+    val snakeYaml        = "2.0"
```
istreeter, line 38: If I run `dependencyTree` in sbt then I see this:

```
[info]   +-com.github.ua-parser:uap-java:1.5.4
[info]   | +-org.apache.commons:commons-collections4:4.4
[info]   | +-org.yaml:snakeyaml:1.33 (evicted by: 2.0)
[info]   | +-org.yaml:snakeyaml:2.0
[info]   | 
```

Is this deliberate to evict 1.33 with a new major version, and are you sure there is no binary incompatibility?

modules/nsq/src/main/scala/com/snowplowanalytics/snowplow/enrich/nsq/GcsClient.scala
istreeter: My mistake, I was looking at the wrong dependencies.

project/Dependencies.scala
benjben: I also don't like that. I did it like in [this OSS contribution](https://github.com/snowplow/enrich/issues/761).

We do have unit tests that use this library and they pass, but I don't know if the eviction happens for the unit tests or only when we build the artifact. do you know ?

Let me check the vulnerabilities with `1.33`.

istreeter: > I don't know if the eviction happens for the unit tests or only when we build the artifact. do you know ?

Eviction happens for unit tests too.  So it was probably OK.

Even so, if you're finding that 1.33 has no vulnerabilities, then it seems safer to use 1.33.

Thanks for checking!

istreeter (approved):

stanch (changes requested):
Could you revert the change of `cloud` telemetry field to `Azure`? I.e. leave it `null` as it was before. I think we will need a more robust mechanism for this, otherwise we will wrongly label a lot of Kafka users as Azure users

modules/kafka/src/main/scala/com.snowplowanalytics.snowplow.enrich.kafka/Main.scala
benjben: I reverted the change. We should address this when we add support for `BlobStorage` to `enrich-kafka`

PR #814, opened 2023-09-06
Add eventbride module

AlexITC:
The Enrich EventBridge module streams Snowplow enriched data in JSON (with the original base64 TSV payload) to AWS EventBridge, which enables routing events based on any parameter on the payload while retaining compatibility for use with other Snowplow components downstream (S3 loader for example).

This contribution is approved for release and license to recipients of Snowplow Enrich only under the Apache License, version 2.0.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -55,6 +55,11 @@ jobs:
       env:
         AWS_ACCESS_KEY_ID: foo
         AWS_SECRET_ACCESS_KEY: bar
+    - name: Run integration tests for enrich-eventbridge
+      run: sbt "project eventbridgeDistroless" IntegrationTest/test
+      env:
+        AWS_ACCESS_KEY_ID: foo
+        AWS_SECRET_ACCESS_KEY: bar
     - name: Run integration tests for enrich-kafka
       run: |
         sbt "project kafka" "docker:publishLocal"
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ lazy val root = project.in(file("."))
   .settings(projectSettings)
   .settings(compilerSettings)
   .settings(resolverSettings)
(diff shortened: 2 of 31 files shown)


AlexITC:
I'm attaching a screenshot with the tests execution:

tests-outcome

stanch:
Hi @AlexITC,

Thanks for this! I’ll recap a few points from the discourse discussion:

> * Our general approach is to keep each application’s functionality to the minimum. Collector collects, Enrich enriches and validates, Loaders load to warehouses, and Snowbridge forwards events to other systems and streams. Adding more flexibility to Enrich seems to go against that.
> * As we expand into Azure, the plan is to rely less and less on cloud-specific components and services (Kinesis, Pub/Sub, EMR, …) and replace them with something cloud-agnostic. Again, an integration with EventBridge would be a step in the opposite direction, and would only cater to AWS users.
> 
> For these reasons, I would prefer the EventBridge integration to be a separate module rather than be embedded in Enrich. Happy to discuss technical details in the PR!

Additionally, could you elaborate on what you mean by this?
> This contribution is approved for release and license to recipients of Snowplow Enrich only under the Apache License, version 2.0.

As [announced](https://snowplow.io/blog/license-changes-for-snowplow-open-source/), the next major version of Enrich — and this feature would warrant a major version bump — is set to use the Snowplow Community License.

alexanderdean:
Always happy to see community contributions but a bit confused by this one. Why wouldn't this just be a standalone microservice that reads from the enriched event stream (i.e. downstream of Enrich) and writes to EventBridge?

Snowplow pipeline is designed to be modular...

AlexITC:
For some reason I missed replying to this thread earlier.

> Why wouldn't this just be a standalone microservice that reads from the enriched event stream (i.e. downstream of Enrich) and writes to EventBridge?

The main reason is that the new modules shares the same core than enrich, creating a microservice requires copying most of enrich code. Have you considered publishing the modules to maven central?

PR #819, opened 2023-10-04
Additional logging

benjben:
Additional DEBUG logs:
- Log the size of the biggest event of a chunk (an event being a collector payload that can contain several actual events)
- Log how many events a collector payload contains
For each event (identified by `event_id`):
- Log when the contexts are going to be validated
- Log when the contexts have been validated
- Log when the enrichments are going to be executed
- Log when the enrichments have been executed
- Log when the contexts added by the enrichments are going to be validated
- Log when the contexts added by the enrichments have been validated


peel (approved):
This is going to be massively costly in a high volume pipeline, but let's try it.

benjben:
Yeah way too costly. I removed the lines that are per event to keep only the ones that are by chunk and by collector payload

PR #820, opened 2023-10-05
Emit bad row if enriching times out

benjben:


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -130,6 +144,7 @@ object Enrich {
       for {
         etlTstamp <- Clock[F].realTime(TimeUnit.MILLISECONDS).map(millis => new DateTime(millis))
         registry <- enrichRegistry
+        _ <- if (payload.toString.contains("ben-testing")) Timer[F].sleep(2.minutes) else Sync[F].unit
```
peel, line 147: Don't think the condition gets a lot of `true`s in prod :)

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -85,10 +86,23 @@ object Enrich {
...
             begin <- Clock[F].realTime(TimeUnit.MILLISECONDS)
-            result <-
-              env.semaphore.withPermit(
-                chunk.toList.map { case (orig, bytes) => enrich(bytes).map((orig, _)) }.parSequenceN(env.streamsSettings.concurrency.enrich)
-              )
+            result <- env.semaphore.withPermit(
+                        chunk.toList
+                          .map {
+                            case (orig, bytes) =>
+                              val timeout = 1.minute
```
peel, line 93: If we notice that the timeout needs a change, we'd make it configurable. May not worth before.

peel (changes requested):

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: Haha yeah it was to test on my pipeline, I forgot to remove it!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
benjben: Definitely!

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/Enrich.scala
```diff
@@ -85,10 +86,23 @@ object Enrich {
...
+                              val badRow = genericBadRow(
```
spenes, line 94: It seems like we are creating this bad row for every event irrespective from timeout. Might not need to change it immediately but it would be good to change it in case we want to put this feature to Enrich permanently.

benjben: Yes good point ! We'll make it better if we decide to keep this feature (we likely will)

PR #830, opened 2023-11-16
enrich-kafka: add blob storage support

spenes:
Main aim of this PR is adding Azure Blob Storage client to enrich-kafka. Since enrich-kafka is cloud agnostic, I've added support for GCS and S3 as well. Also, I've integrated Azure Blob Storage client to enrich-nsq.

While doing these changes, I've made a bit refactoring. Firstly, I've moved blob storage clients to their own module. Secondly, I've moved config tests to their own modules. With this way, tests pick up modules' `application.conf` files.


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
```diff
@@ -486,6 +486,24 @@ object io {
       deriveConfiguredEncoder[GcpUserAgent]
   }
 
+  case class BlobStorageClients(
+    gcs: Boolean,
+    s3: Boolean,
```
istreeter, line 491: Imagine we removed these booleans from the configuration.  What would be the disadvantage?

I guess you are trying to avoid creating clients if they will never be needed.  Especially as each client consumes JVM resources (threads, memory, etc).

If that is what you are trying to avoid....  Is it possible to only create the client **if** the `canDownload` matches.

For example, with `AzureStorageclient` you could move creating the store into the `download` method:

```scala
new Client[F] {
  def canDownload(uri: URI): Boolean = ???

  def download(uri: URI): Stream[F, Byte] =
    Stream.resource(createStore(storageName)).flatMap { store =>
      ???
    }
}
```

istreeter (commented):
I would be OK to approve this.  I'm just trying to probe if we can remove any complexity by doing it slightly differently.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
spenes: Sounds like a good idea! I will start to make necessary changes.

istreeter (approved):

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
istreeter: I'm OK with that.  Thanks for considering it.

PR #836, opened 2023-12-01
Upgrade to Cats Effect 3 ecosystem

benjben:
Main changes:
- `Blocker` removed
- `RegistryLookup` passed explicitly rather than implicitly
- There is still a cached thread pool created manually and used only for IP lookup enrichment (waiting for changes in the lib)


modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/EnrichSpec.scala
```diff
@@ -93,25 +96,29 @@ class EnrichSpec extends Specification with CatsIO with ScalaCheck {
...
+              )(
+                payload
+              )
+              .map(normalizeResult)
+              .map {
+                case List(Validated.Valid(e)) => e.event must beSome("page_view")
+                case other => ko(s"Expected one valid event, got $other")
+              }
+          }
+          .unsafeRunSync()
```
istreeter, line 121: The old version didn't need this `unsafeRunsync()`.  How come you needed to add it?

istreeter (approved):
Looks great.  Please remind me, are we hoping the `blockingEC` will be removed in the final final version?

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/CleanCancellation.scala
```diff
@@ -0,0 +1,115 @@
...
+import cats.implicits._
+
+import cats.effect.implicits._
+import cats.effect.{Async, Deferred, Sync}
+import cats.effect.kernel.Resource.ExitCase
+import cats.effect.std.Queue
+
+import fs2.{Pipe, Stream}
+
+object CleanCancellation {
```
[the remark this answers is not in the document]
benjben: I took it from here : https://github.com/snowplow-incubator/common-streams/blob/develop/modules/streams-core/src/main/scala/com.snowplowanalytics.snowplow/sources/internal/CleanCancellation.scala
That's a step towards using the lib in Enrich :smiley:

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/EnrichSpec.scala
benjben: Good question, I'm not sure which dependency bump broke this, but there was an implicit missing. I tried using https://github.com/typelevel/scalacheck-effect but it didn't work out of the box and I thought that it was not worth it to spend time on it, although I don't like this `unsafeRunsync()`

benjben:
Thanks for the review @istreeter !

> are we hoping the blockingEC will be removed in the final final version?

As much as I'd love that, I doubt it. There is a JIRA ticket (`PDP-737`) to look at the source code of [scala-maxmind-iplookups](https://github.com/snowplow/scala-maxmind-iplookups) to figure out if it performs IO, but it is not prioritized.

modules/common-fs2/src/test/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/EnrichSpec.scala
istreeter: Ah I see -- it's something to do with the combination of using scalacheck with io.

That's fine.  I'm ok with leaving `unsafeRunSync()`.

benjben: > it's something to do with the combination of using scalacheck with io

Exactly ! But we were already doing this before

PR #840, opened 2023-12-12
Allow passing an object of parameters to the JS enrichment

stanch:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/JavascriptScriptEnrichment.scala
```diff
@@ -54,21 +55,27 @@ object JavascriptScriptEnrichment extends ParseableEnrichment {
...
   private val enrichmentInfo =
     FailureDetails.EnrichmentInformation(schemaKey, "Javascript enrichment").some
 
   private val engine = new ScriptEngineManager(null)
     .getEngineByMimeType("text/javascript")
     .asInstanceOf[ScriptEngine with Invocable with Compilable]
 
-  private val stringified = rawFunction + """
+  private val stringified = rawFunction + s"""
+    var _params = ${params.asJson.noSpaces};
```
istreeter, line 76: Just for fun, I played with whether it's possible to hide the `_params` from the global namespace: #841

PR #844, opened 2023-12-20
Support for multiple Azure blob storage account

spenes:


istreeter (approved):

config/config.kafka.extended.hocon
```diff
@@ -235,8 +235,8 @@
     # Optional. Azure blob storage client configuration.
     # ABS client won't be enabled if it isn't given.
     "azureStorage": {
-      # Azure blob storage account name
-      "storageAccountName": "storageAccount"
+      # Azure blob storage account names
+      "storageAccountNames": ["storageAccount1", "storageAccount2"]
```
istreeter, line 239: This is a non-backward compatible config change, right?

I'm ok with that if the previous config was unreleased / experimental / not-used.  It would be annoying to add a custom decoder to handle the case when it's a simple string -- but did you consider adding the custom decoder?

Should have asked this before I approved it :)

spenes: I did it like this because original feature isn't released yet. It is still in the develop branch. But I can add custom decoder as well. It should be easy change 👍

istreeter: That's perfect if the original is not released yet!

Please don't add a custom decoder!!!  If we can get away without it then I would prefer to keep things as simple as possible.

PR #845, opened 2023-12-27
Add different types of authentication for azure blob storage

pondzix:
This PR makes azure blob storage more flexible regarding authentication methods. Now t's possible to configure multiple storage accounts (like before), but with: 

* no authentication at all. No credentials are used when accessing storage. Useful for public accounts.
* authentication using default credentials [chain](https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable). That's the method Enrich used previously for all configured accounts.
* authentication using SAS token. Token is provided in configuration and NOT generated dynamically in application. 

New configuration looks like: 

```
    "azureStorage": {
      "accounts": [
        # Example public account with no auth
        { "name": "storageAccount1"},

        # Example private account using default auth chain -> https://learn.microsoft.com/en-us/java/api/com.azure.identity.defaultazurecredential?view=azure-java-stable
        { "name": "storageAccount2", "auth": { "type": "default"} },

        # Example private account using SAS token auth
        { "name": "storageAccount3",  "auth": { "type": "sas", "value": "tokenValue"}}
      ]
    }
```

Comparing to the old one:

```
    "azureStorage": {
      # Azure blob storage account names
      "storageAccountNames": ["storageAccount1", "storageAccount2"]
    }
```

### DIFF
--- a/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
+++ b/modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
@@ -20,8 +20,7 @@ import java.util.UUID
 import cats.syntax.either._
 
 import scala.concurrent.duration.{Duration, FiniteDuration}
-
-import _root_.io.circe.{Decoder, DecodingFailure, Encoder}
+import _root_.io.circe.{Codec, Decoder, DecodingFailure, Encoder}
 import _root_.io.circe.generic.extras.semiauto._
 import _root_.io.circe.config.syntax._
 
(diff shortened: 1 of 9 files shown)


modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
```diff
@@ -492,8 +491,47 @@ object io {
...
+    }
+
+    implicit val sasTokenDecoder: Codec[AzureStorage.Account.Auth.SasToken] =
+      deriveConfiguredCodec
+
+    implicit val accountAuthDecoder: Decoder[AzureStorage.Account.Auth] =
+      Decoder.instance { cursor =>
+        val typeCur = cursor.downField("type")
+        typeCur.as[String].map(_.toLowerCase) match {
+          case Right("chain") =>
```
istreeter, line 516: IMO "default" would be a better type name to put in the config file.  Meaning Azure's _default_ credentials chain.

istreeter (approved):

PR #849, opened 2024-01-08
Add mandatory SLULA license acceptance flag

spenes:


benjben (approved):
Looks great :+1:

config/config.kafka.extended.hocon
```diff
@@ -1,4 +1,9 @@
 {
+  # Full license text available in LICENSE.md
+  license {
+    accept = true
+  }
```
istreeter, line 5: This new section stands out as being in a different style from the rest of the file.

To be in-keeping with the rest of the file, it would look like this:

```
"license": {
  "accept": true
}
```

....which is what you've done in the `*.minimal.hocon` examples.

modules/common-fs2/src/main/scala/com/snowplowanalytics/snowplow/enrich/common/fs2/config/io.scala
```diff
@@ -622,4 +622,19 @@ object io {
 
     implicit val encoder: Encoder[Cloud] = Encoder.encodeString.contramap[Cloud](_.toString().toUpperCase())
   }
+
+  case class License(
+    accept: Boolean
+  )
+
+  object License {
+    implicit val licenseDecoder: Decoder[License] = {
```
istreeter, line 631: Because I find it fun to see if there's alternative ways to write the same thing in a different way.  Here's what I came up with:

```scala
    implicit val licenseDecoder: Decoder[License] = {
      val truthy = Set("true", "yes", "on", "1")
      Decoder[String]
        .map(_.toLowerCase)
        .map(truthy(_))
        .or(Decoder[Boolean])
        .map(License(_))
        .at("accept")
    }
```
or

```scala
    implicit val licenseDecoder: Decoder[License] = {
      val truthy = Set("true", "yes", "on", "1")
      Decoder[String]
        .either(Decoder[Boolean])
        .at("accept")
        .map {
          case Left(s) => License(truthy(s.toLowerCase))
          case Right(b) => License(b)
        }
    }
```

or what you did, which is completely fine and is the same as the collector

PR #857, opened 2024-01-25
Bump libs to address security vulnerabilities

benjben:


.github/workflows/release.yml
```diff
@@ -149,18 +153,24 @@ jobs:
...
         command: monitor
       env:
         SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
+    - name: Run Snyk to check for vulnerabilities (-rc)
+      uses: snyk/actions/docker@master
+      if: ${{ ! matrix.final_release }}
+      with:
+        image: "${{ steps.packageName.outputs.package_name }}:${{ steps.ver.outputs.tag }}-distroless"
+        args: "--app-vulns"
+        command: monitor
```
istreeter, line 173: Is the intention for this to go to our non-prod dashboard?

benjben: Yeah, or just display it in the Github action, but that [failed](https://github.com/snowplow/enrich/actions/runs/7668099290/job/20899268003)

istreeter: How does it work though?  I assumed our `SNYK_TOKEN` is scoped only for one organization?  Or is that a wrong assumption -- maybe it has permissions to push to both organizations?

Should we specify the `--org=something` so we explicitly tell it where to send the results?

benjben: You're right, it is scoped for `data-processing-new`. I removed it

`for queue in kinesis pubsub kafka nsq; do sbt "project ${queue}Distroless; set version := \"latest\"; Docker / publishLocal"; snyk container test --app-vulns snowplow/snowplow-enrich-$queue:latest-distroless; done` FTW!

colmsnowplow (approved):
LGTM

PR #864, opened 2024-01-31
enrich-kafka: authenticate with Event Hubs using OAuth2

spenes:
We've added same feature on collector as well: https://github.com/snowplow/stream-collector/pull/402


istreeter (approved):

spenes:
Merged to develop, closing.

PR #874, opened 2024-02-14
Add configuration for incomplete events sink

benjben:


config/config.file.extended.hocon
```diff
@@ -31,6 +31,14 @@
       "file": "/var/bad"
       "maxBytes": 1000000
     }
+
+    # Incomplete events output
+    "incomplete": {
```
colmsnowplow, line 36: I assume simply omitting this key disables the feature yeah?

benjben: Yes! Similarly too pii events

colmsnowplow: Super. Love it.

PR #878, opened 2024-02-20
Add integration test for incomplete sink

benjben:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -123,6 +123,10 @@ object EnrichmentManager {
     val iorT = for {
       _ <- IorT.fromIor[F](setupEnrichedEvent(raw, enrichedEvent, etlTstamp, processor))
       extract <- IgluUtils.extractAndValidateInputJsons(enrichedEvent, client, registryLookup)
+      _ = {
+        enrichedEvent.contexts = ME.formatContexts(extract.contexts).orNull
+        enrichedEvent.unstruct_event = ME.formatUnstructEvent(extract.unstructEvent).orNull
```
benjben, line 128: We were missing these very important 2 lines

istreeter (approved):

PR #883, opened 2024-02-29
Switch bad row type from EnrichmentFailures to SchemaViolations for s…

benjben:
There are 3 places that currently produce an `EnrichmentFailures` bad row whereas the more appropriate type is `SchemaViolations`:

1. When the input fields of the HTTP requests are mapped to the atomic event.
2. When the enrichments contexts get validated.
3. When the atomic fields lengths get validated.

For 1 and 3, the error should be mapped into an Iglu `ValidationError` with the atomic schema referenced.

Before this change, if there was any error in the mapping of the atomic fields, a bad row would get emitted right away and we would not try to validate the entities and unstruct event. Now all these errors are wrapped inside a same `SchemaViolations` bad row.

Likewise, before this change when an enrichment context was invalid, we were emitting a bad row right away and not checking the lengths of the atomic fields. Now all these errors are wrapped inside a same `SchemaViolations` bad row.

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFieldsLengthValidator.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFieldsLengthValidator.scala
@@ -14,14 +14,14 @@ import org.slf4j.LoggerFactory
 
 import cats.Monad
 import cats.data.Validated.{Invalid, Valid}
-import cats.data.{NonEmptyList, ValidatedNel}
+import cats.data.NonEmptyList
 
 import cats.implicits._
 
-import com.snowplowanalytics.snowplow.badrows.FailureDetails.EnrichmentFailure
-import com.snowplowanalytics.snowplow.badrows.{BadRow, FailureDetails, Processor}
+import com.snowplowanalytics.iglu.client.validator.ValidatorReport
 
-import com.snowplowanalytics.snowplow.enrich.common.adapters.RawEvent
+import com.snowplowanalytics.snowplow.badrows.FailureDetails
+
 import com.snowplowanalytics.snowplow.enrich.common.enrichments.AtomicFields.LimitedAtomicField
 import com.snowplowanalytics.snowplow.enrich.common.outputs.EnrichedEvent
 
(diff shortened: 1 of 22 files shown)


istreeter (commented):
Looks good @benjben!

Regarding my suggested improvement... I have not assessed this needs changing in more places than just the place I highlighted.  If you assess that it stretches far beyond just the bits of code that you changed here, then maybe it needs to be its own separate commit.

(Or, of course, reject my idea outright)

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFieldsLengthValidator.scala
```diff
@@ -35,66 +32,46 @@ object AtomicFieldsLengthValidator {
...
         Monad[F].pure(Right(()))
     }
 
   private def validateField(
-    event: EnrichedEvent
-  )(
+    event: EnrichedEvent,
     atomicField: LimitedAtomicField
-  ): ValidatedNel[String, Unit] = {
+  ): Either[AtomicError, Unit] = {
```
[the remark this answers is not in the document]
benjben: I like this idea! I'll remove `AtomicError` and use `ValidatorReport` everywhere :+1:

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -84,19 +93,44 @@ object EnrichmentManager {
...
+    etlTstamp: DateTime,
+    processor: Processor,
+    client: IgluCirceClient[F],
+    registryLookup: RegistryLookup[F]
+  ): EitherT[F, BadRow, IgluUtils.EventExtractResult] =
+    EitherT {
+      val setup = setupEnrichedEvent(raw, enrichedEvent, etlTstamp, processor).toValidatedNel
+      IgluUtils
+        .extractAndValidateInputJsons(enrichedEvent, client, registryLookup)
+        .map((setup, _).mapN((_, extract) => extract))
```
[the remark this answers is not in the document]
benjben: Great point !

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -312,34 +313,35 @@ object EnrichmentManager {
...
-    val collectorTstamp = setCollectorTstamp(e, raw.context.timestamp).toValidatedNel
-    // Map/validate/transform input fields to enriched event fields
-    val transformed = Transform.transform(raw, e)
-
-    (collectorTstamp |+| transformed)
-      .leftMap(AtomicFields.atomicErrorsToSchemaViolation)
-      .toEither
-  }
+  ): F[Either[FailureDetails.SchemaViolation, Unit]] =
+    Monad[F].pure {
```
istreeter, line 322: Is there a reason you went for `Monad[F].pure` instead of `Sync[F].delay`?  It's fine if the answer is because that would be too much of a change, to introduce a `Sync` into this part of the code.

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
```diff
@@ -115,20 +116,20 @@ object EnrichmentManager {
...
-              RawEvent.toRawEvent(raw),
-              processor
-            )
-          }.toEither
+      for {
+        setup <- setupEnrichedEvent[F](raw, enrichedEvent, etlTstamp, processor).map(_.toValidatedNel)
+        iglu <- IgluUtils.extractAndValidateInputJsons(enrichedEvent, client, registryLookup)
+      } yield (setup, iglu)
+        .mapN((_, extract) => extract)
+        .leftMap { violations =>
```
istreeter, line 124: How about

```scala
} yield (iglu <* setup)
  .leftMap { violations =>
```

or 

```scala
} yield iglu
  .productL(setup)
  .leftMap { violations =>
```

Personally I find those easier to understand than `.mapN((_, extract) => extract)` but it's fine if you disagree!

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
benjben: Indeed it's because only `Monad` was available here and I didn't see a point in introducing `Sync`. What would the advantage be? It's not a big change to introduce it (we'll very likely end up needing `Sync` at some point anyway)

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
benjben: Oh yeah you had already told me and I forgot, sorry ! Agree that it looks better

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/EnrichmentManager.scala
benjben: Someone who does all that would really try to break Enrich on purpose :joy: Tests would break anyway, but yeah let's add a bit of FP good practices, even though we have this disgusting mutable event, I'll introduce `Sync`

benjben: Now I'm really tempted to wrap all the places of Enrich that mutate the event inside `Sync[F].delay()`, but making Enrich more functional is outside of this task

benjben:
Hey @istreeter , I addressed all your feedback, please have a look

istreeter (approved):

PR #884, opened 2024-02-29
Incomplete events failure payloads contd

oguzhanunlu:
prepared based on schemas of schema_violation and enrichment_failures

### DIFF
--- a/modules/common/src/test/resources/iglu-schemas/schemas/com.acme/malformed_schema/jsonschema/1-0-0
+++ b/modules/common/src/test/resources/iglu-schemas/schemas/com.acme/malformed_schema/jsonschema/1-0-0
@@ -1,0 +1,24 @@
+{
+  "$schema": "http://iglucentral.com/schemas/com.snowplowanalytics.self-desc/schema/jsonschema/1-0-0#",
+  "description": "Schema for acme stuff",
+  "self": {
+    "vendor": "com.acme",
+    "name": "malformed_schema",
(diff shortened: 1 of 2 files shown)


benjben (commented):
I think that you cover all the cases of `SchemaViolations` :+1: 
- [ResolutionError](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/ClientError.scala#L39)
  - [LookupHistory](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/resolver/LookupHistory.scala#L34)
- [ValidationError](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/ClientError.scala#L45)
  - [ValidatorError](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorError.scala#L22)
    - [ValidatorReport](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorReport.scala#L19)
    - [SchemaIssue](https://github.com/snowplow/iglu-scala-client/blob/master/modules/data/src/main/scala/com.snowplowanalytics.iglu/client/validator/ValidatorError.scala#L40)

PR #890, opened 2024-04-08
Remove too long atomic fields

oguzhanunlu:


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFields.scala
```diff
@@ -32,97 +32,98 @@ object AtomicFields {
...
-      AtomicField(name = "domain_sessionid",   _.domain_sessionid   ),
-      AtomicField(name = "event_vendor",       _.event_vendor       ),
-      AtomicField(name = "event_name",         _.event_name         ),
-      AtomicField(name = "event_format",       _.event_format       ),
-      AtomicField(name = "event_version",      _.event_version      ),
-      AtomicField(name = "event_fingerprint",  _.event_fingerprint  ),
+      AtomicField(name = "app_id", _.app_id, _.app_id = null),
+      AtomicField(name = "platform", _.platform, _.platform = null),
+      AtomicField(name = "event", _.event, _.event = null),
+      AtomicField(name = "event_id", _.event_id, _.event_id = null),
```
istreeter, line 46: For some loaders `event_id` is a a required field in the table.  For example, see [here](https://github.com/snowplow-incubator/snowflake-loader/blob/0.2.2/modules/core/src/main/scala/com.snowplowanalytics.snowplow.snowflake/processing/TableManager.scala#L147) how we create the snowflake events table.

Same is true of `v_collector` and `v_etl`, although I cannot tell you why it is this way!!

I'm fine with setting fields to null, and I understand how this helps with the wider design of incomplete events.  But be aware this makes it technically possible that enrich can emit events which cannot be loaded.

cc @colmsnowplow

istreeter (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFields.scala
oguzhanunlu: thanks for the heads up @istreeter ! That makes me think if we should
- keep a list of atomic field names to be excluded from this check or
- take prefix of the field as much as its' limit allow or
- let it be and accept the possibility

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFields.scala
```diff
@@ -43,11 +43,11 @@ object AtomicFields {
       AtomicField(name = "app_id", _.app_id, _.app_id = null),
       AtomicField(name = "platform", _.platform, _.platform = null),
       AtomicField(name = "event", _.event, _.event = null),
-      AtomicField(name = "event_id", _.event_id, _.event_id = null),
+      AtomicField(name = "event_id", _.event_id, _ => ()),
```
benjben, line 46:
```suggestion
      AtomicField(name = "event_id", _.event_id, _ => ()), // required in loading
```

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFields.scala
```diff
@@ -43,11 +43,11 @@ object AtomicFields {
...
       AtomicField(name = "name_tracker", _.name_tracker, _.name_tracker = null),
       AtomicField(name = "v_tracker", _.v_tracker, _.v_tracker = null),
-      AtomicField(name = "v_collector", _.v_collector, _.v_collector = null),
-      AtomicField(name = "v_etl", _.v_etl, _.v_etl = null),
+      AtomicField(name = "v_collector", _.v_collector, _ => ()),
+      AtomicField(name = "v_etl", _.v_etl, _ => ()),
```
benjben, line 50:
```suggestion
      AtomicField(name = "v_collector", _.v_collector, _ => ()), // required in loading
      AtomicField(name = "v_etl", _.v_etl, _ => ()), // required in loading
```

benjben (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFields.scala
oguzhanunlu: add clarifying comment for v_collector and v_etl

PR #891, opened 2024-04-19
Set data field of incomplete event failure entity to SDJ's data field only

spenes:
At the moment, we set all self describing json to data field of incomplete event failure entity. However, failure entity has separate schema field and therefore we don't need to set all SDJ to data field. It is enough to use only data field of SDJ.


benjben (approved):
:+1:

PR #892, opened 2024-04-30
Don't pass field value to ValidatorReport if validation fails

spenes:
In 4.2.0, we've started to use `SchemaViolations` instead of `EnrichmentFailures` with some errors. While doing this change, we've used `ValidatorReport` to wrap atomic field validation errors and we've passed field values as well to `ValidatorReport`. However, field values shouldn't be passed to `ValidatorReport`. This PR makes necessary changes to remove those values from `ValidatorReport`.


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/AtomicFieldsLengthValidator.scala
```diff
@@ -53,14 +52,13 @@ object AtomicFieldsLengthValidator {
   private def validateField(
     event: EnrichedEvent,
     atomicField: LimitedAtomicField
-  ): Either[ValidatorReport, Unit] = {
+  ): Either[AtomicFieldValidationError, Unit] = {
```
benjben, line 55:
```suggestion
  ): Either[FieldLengthError, Unit] = {
```
once we have 
```
sealed trait AtomicError
case class FieldLengthError(...) extends AtomicError
```
(other comment)

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/utils/MapTransformer.scala
```diff
@@ -66,7 +64,7 @@ object MapTransformer {
   type Field = String
 
   // A transformation takes a Key and Value and returns either a failure or anything
-  type TransformFunc = Function2[Key, Value, Either[ValidatorReport, _]]
+  type TransformFunc = Function2[Key, Value, Either[AtomicFieldValidationError, _]]
```
benjben, line 67:
```suggestion
  type TransformFunc = Function2[Key, Value, Either[ParseError, _]]
```
(see other comments)

benjben (changes requested):

benjben (approved):
Looks great @spenes !

PR #898, opened 2024-05-22
Replace mysql client with mariadb client

oguzhanunlu:
Jira ref: PDP-1206

### DIFF
--- a/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
+++ b/modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
@@ -56,13 +56,12 @@ object Rdbms {
     database: String
   ) extends Rdbms {
 
-    val driver: Class[_] = Class.forName("com.mysql.cj.jdbc.Driver") // Load class
+    val driver: Class[_] = Class.forName("org.mariadb.jdbc.Driver") // Load class
+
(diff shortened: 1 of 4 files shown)


modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
```diff
@@ -56,13 +56,12 @@ object Rdbms {
     database: String
   ) extends Rdbms {
 
-    val driver: Class[_] = Class.forName("com.mysql.cj.jdbc.Driver") // Load class
+    val driver: Class[_] = Class.forName("org.mariadb.jdbc.Driver") // Load class
+
+    val sslModeParam: String = if (sslMode) "&sslMode=trust" else "&sslMode=disable&allowPublicKeyRetrieval=true"
```
istreeter, line 61: I agree that "trust" the sslMode to pick to be compatible with previous versions of Enrich.  But interestingly [in these docs](https://mariadb.com/docs/server/connect/programming-languages/java/tls/) it is described as "not safe for production applications."

istreeter (approved):

modules/common/src/main/scala/com.snowplowanalytics.snowplow.enrich/common/enrichments/registry/sqlquery/Rdbms.scala
oguzhanunlu: Thanks @istreeter ! Higher levels require ssl cert setup/config on client and database side. From enrich's perspective, we can let user configure cert path and raise ssl mode to e.g. `verify-ca`. Proceeding with current behavior at the moment.

PR #899, opened 2024-06-12
🚧 Validate cross navigation for 4.0.0 enrich

peel:
The purpose of this test is to verify that enric version prior to 4.1.0 will not break with additional cross_nagivation_config enrichment.

The test adds `cross_navigation_config.json` to Kafka integration test to prove the safe behavior.


peel:
Proved as per spec.

PR #902, opened 2024-06-24
Pass request headers to the JS enrichment (close #901)

### LINKED ISSUE #901: Provide request headers to the JS enrichment
This allows custom logic, such as extracting JWTs and other API tokens to be used by the enrichment.

It is already possible to enable the HTTP Header enrichment and then access the entity in JS, however then there is no way to remove the sensitive header values afterwards. Scrubbing them with the PII enrichment is a workaround, but that adds more data to be stored in the warehouse, for no reason.

Accessing the headers directly is a much more streamlined approach.

See also #66 which would be solved by this.

stanch:
Pass a raw list of headers as the third argument (after event and parameters) to the JavaScript enrichment function.

Example usage:

```js
function process(event, params, headers) {
  for (header of headers) {
    const jwt = header.match(/X-JWT:(.+)/i)
    if (jwt) {
      doSomething(jwt[1].trim())
    }
  }
}
```


benjben (approved):
Looks good :+1:

PR #906, opened 2024-12-19
Remove GH workflows release script

AlexBenny:


AlexBenny:
Closing as we prefer a different solution

